PR #30, opened 2015-07-25
Fixing MuxServerSocketChannel when binding to "anyLocalAddress" and attempting to reuse socket by binding a particular interface

robertoandrade:
As discussed in http://lists.jitsi.org/pipermail/dev/2015-July/024766.html


lyubomir:
Thank you very much for the contribution. I'd like to gather opinions on the details here:

Let's say "any local address" covers addresses A and B, we have (1) one MuxServerSocketChannel bound to "any local address" and (2) another one bound to address A.

The way you've done it, if a connection arrives on address B and it matches the filter of (2), (2) will accept it even though it was bound to address A only.

Do we want this behavior? Or do we actually want to make sure that (2) does not accept the connection that arrived on address B because it was bound to address A only?

lyubomir:
I and a couple of colleagues had an offline discussion so I wanted to bring you up to speed with our latest view on the subject. We will "want to make sure that (2) does not accept the connection that arrived on address B because it was bound to address A only". Since this behavior includes your proposed modifications, we'll be happy to accept them now and extend them later. As stated at https://jitsi.org/Documentation/FAQ#patch, we will need you to sign our contributor agreement as either an individual (https://jitsi.org/icla) or a corporation (https://jitsi.org/ccla) before we can merge your pull request. Thank you very much!

robertoandrade:
Makes sense. Just submitted the signature via DocuSign.

robertoandrade:
Any idea when this is going to make it to the latest build of JVB?

PR #34, opened 2015-10-13
Remove some hacks that call setUseCandidateReceived

dpocock:
I'm not sure if this is everything that has to be fixed, but it resolved the issue for me.

I feel that logging improvements would also help with this type of issue.  The current log messages are ambiguous.  Because the code is multi-threaded, the order of the log messages can also be confusing.  Given the nature of the protocol, it would be helpful for each log message to be longer, showing all the relevant details of the candidate pair and check it relates to, on a single log line.

### DIFF
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -1515,29 +1515,6 @@ public class Agent
             //we already know about the remote address so we only need to
             //trigger a check for the existing pair
 
-            if(!isControlling())
-            {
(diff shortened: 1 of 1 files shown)


dpocock:
Discussed in http://lists.jitsi.org/pipermail/dev/2015-October/025372.html

ibauersachs:
Could you please sign the new [Atlassian/Apache CLA](https://jitsi.org/icla)?

dpocock:
did you receive confirmation that the CLA was completed?

ibauersachs:
No, I'll try to ask Emil today.

PR #38, opened 2015-10-27
Mostly fixes and adjustments to support my turnserver updates

mondain:
Fixed pseudotcp recv window test to not fail when window scaling is not supported. Added builders to support selected immutable instance use by external projects like turnserver. Modified MessageFactory to support a couple of turnserver method calls.

Added an eclipse code format file to make jitsi code format easier to apply.

### DIFF
--- a/code-format-jitsi.xml
+++ b/code-format-jitsi.xml
@@ -1,0 +1,295 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<profiles version="12">
+<profile kind="CodeFormatterProfile" name="Jitsi" version="12">
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_ellipsis" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_enum_declarations" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_new_line_in_empty_annotation_declaration" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_before_comma_in_allocation_expression" value="do not insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_before_at_in_annotation_type_declaration" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.comment.new_lines_at_block_boundaries" value="true"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_constructor_declaration_parameters" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.comment.insert_new_line_for_parameter" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_new_line_after_annotation_on_package" value="insert"/>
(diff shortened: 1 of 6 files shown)


ibauersachs:
As with the Turnserver PR, please create separate PRs for individual changes.
- The formatter shouldn't be in the repo (at most it can go into e.g. Jitsi-Universe)
- Please don't reformat existing code
- And there are style issues

mondain:
The formatter was dropped in hoping that it would be adjusted to properly support the groups code style; I will however do a PR for universe instead. 
There may have been some unintentional reformatting, but this was after noticing that some lines exceeded lengths enforced by the team when I submit a PR; however it does look like some of the line indents are larger than 4 spaces. If there is an actual style guide, it would be helpful to know what the rules are before submitting the PR.
As for separate PR's, there are only 6 files here; minus the formatter and unit test fix, the others are required to allow turnserver to work.

ibauersachs:
The problem with the formatter template is
- it hasn't been used by everyone
- the one others are using from Idea is slightly different
- some do manual formatting by hand.
This all ends up with following the [convention](https://jitsi.org/Documentation/CodeConvention) as good as possible, and only using the (any) formatter for new code. Certainly not for an entire existing file.

At first glance there were a lot of space-to-tab conversions, which is wrong anyway.

Btw, your pull requests should come from a branch, not master. That way you can update/rebase your branch against your master, which in turn is fast-forward updatable.

mondain:
I've been doing it by hand and maybe its the OCD, but it looked odd having a formatted section surrounded by incorrectly formatted code; I'll stick with only my section from now on. Thanks for the style link, its great to know the rules.
As for branches, I don't have write access so I use a fork and if I'm not mistaken, rebase will work just fine; I'm no git expert.

mondain:
I'm closing this in lieu of more focused an compliant partitioned requests.

ibauersachs:
You'd create the branch in your own repo, not here. See [How to GitHub](https://gun.io/blog/how-to-github-fork-branch-and-pull-request/) for a good explanation.

mondain:
Yes, that I know how to do, but it adds another level of complexity to submission of a PR to the main line project; there is no benefit to forking the main project repo and then creating a branch on the fork. The fork itself is essentially a branch, so I'd be branching a branch.

mondain:
@ibauersachs  I just ran into the line endings issue you spoke of, which I didn't notice or was not alerted of in eclipse previously. The style guide isn't specific here, so which line ending should be selected? The PseudoTcpTestRecvWindow.java has a mix of unix and windows; one or the other should be chosen; so which should I go with?

ibauersachs:
The repo has to be LF (i.e. Unix). But your local file can be anything and actually depends on what OS you develop.
Please make sure that the line endings conversion is the **only** change in that commit. You can include the commit in a PR with other commits though.

mondain:
Thank you for getting right back to me on this; for the record I dev in Ubuntu most of the time.

mondain:
Not sure how many of these I'll get submitted today, but here's the first one for reference:
https://github.com/jitsi/ice4j/pull/39

PR #39, opened 2015-10-28
Fixed setReceiveWindowSize test to allow it to pass when the current …

mondain:
…system does not support window scaling. Also modified the source to use Unix line endings instead of mixed.

Lines 393 to 401 are the only "actual" code modification here.


ibauersachs:
As mentioned in #38, please create two commits:
- one for the line endings cleanup
- one for the actual code change

mondain:
Ok, so at this point I get the feeling that you don't want these commits. There seems to be an inordinate amount of hoops to jump through each time I submit something. If I must, I'll just keep all the fixes and mods to myself and go on about my merry way without helping the community; this is not what I want to do, but I'm sorry, this is getting a little crazy. My notes were specific, I got authorization about the LF and you still want more for a single file PR with just a few actual mods.

ibauersachs:
Paul, I'm sorry, but we need to keep these things clean. Crap happened when these files ended up with non-LF line endings in the repository (SVN back at the time which didn't care), and then again when it wasn't cleaned up during the migration to Git (because someone used the Google Code export button instead of Daniel Pocock's nice and clean export).

I'll clean up the line endings in all files. Then, if you still want to contribute, which I would welcome, you can work on a clean file.

ibauersachs:
Line endings done in da2194444c382b53f1b880389662a60bbb47e8fb.

mondain:
Thanks for taking care of that quickly and I guess I'm just feeling some frustration although I'm sure I'd be at square one without the work you guys have done. Might I ask that you fix the line endings on turnserver too? I haven't changed any of that back yet, this morning.

mondain:
I fixed the PR, here's the new one https://github.com/jitsi/ice4j/pull/40
Also for anyone else reading this, an excellent answer on SO for keeping a fork in-sync with an upstream master: http://stackoverflow.com/a/7244456/127938

PR #40, opened 2015-10-28
Fixed to prevent failure with systems that dont support window scaling.

mondain:
This test was failing previously if your OS did not support window scaling.

### DIFF
--- a/src/test/java/org/ice4j/pseudotcp/PseudoTcpTestRecvWindow.java
+++ b/src/test/java/org/ice4j/pseudotcp/PseudoTcpTestRecvWindow.java
@@ -390,8 +390,15 @@ public class PseudoTcpTestRecvWindow extends PseudoTcpTestBase
         test.setOptNagling(false);
         test.setOptAckDelay(0);
         int wndSize = 300000;
+        // if window scaling is not supported by either local or remote, use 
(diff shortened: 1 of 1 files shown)


ibauersachs:
Hum, I hate to be picky again, but: this pull request should only be about the changes to PseudoTcpTestRecvWindow in commit 847f9a6 (which is fine and good to merge).

I'm not sure if you intentionally added the other four commits to this PR, but I'd guess this happened accidentally because you created the PR from your master branch and then pushed new commits to it - which updates the PR too.

Some comment on the other commits:
- Builder pattern: okay, but please remove the turnserver comment from Javadoc, document the parameters and check the indents. There are some tabs in there.
- Line endings/tab spaces: please squash these two, and they will likely not be necessary at all
- Reformatting: please only do that on actual violation of the convention (there are a few of those). The reason: it makes reading the history hard. If you want to find out why a certain piece of code is like it is or why it was changed (by using git blame), you'd only find the reformat commit.

mondain:
It was not intentional to group them, I'll have to figure out how to cut them out.

mondain:
ok fixed it with

```
git reset --hard <old-commit-id>
git push -f origin HEAD
```

PR #54, opened 2016-03-23
Packet queue

bgrozev:


src/main/java/org/ice4j/util/PacketQueue.java
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    /**
+     * The underlying {@link Queue} which holds packets.
+     */
+    private final Queue<T> queue;
+
+    /**
+     * Whether this {@link PacketQueue} should store the {@code byte[]} or
+     * {@code T} instances added to it via on of the {@code add} methods (of
```
paweldomas: you meant 'via one of the add methods', right ? and "(if false)"

src/main/java/org/ice4j/util/QueueStatistics.java
```diff
@@ -0,0 +1,113 @@
+package org.ice4j.util;
```
paweldomas: missing license header

src/main/java/org/ice4j/util/PacketQueue.java
```diff
@@ -0,0 +1,524 @@
...
+     * Get an unused {@link byte[]} instance with length at least {@code len}.
+     * @param len the minimum length of the returned instance.
+     * @return a {@link byte[]} instance with length at least {@code len}.
+     */
+    private byte[] getByteArray(int len)
+    {
+        if (cache != null)
+        {
+            byte[] buf = cache.poll();
+            if (buf != null && buf.length >= len)
```
paweldomas: is it true that byte[1] in cache queue will block caching for all future packets of length > 1?
we consider byte[] to be unused once it is dropped from the queue ? I do not really grasp the concept, would appreciate some more explanation in javadocs on how the caching works

src/main/java/org/ice4j/util/PacketQueue.java
bgrozev: Fixed.

src/main/java/org/ice4j/util/PacketQueue.java
bgrozev: The rationale is that if the cache contains a byte[] with insufficient length, it is released rather than returned to the cache because we are likely to run into the same situation later on.
Anyway, I have removed the whole cache thing, because it isn't required right now, and the implementation wasn't complete.

src/main/java/org/ice4j/util/QueueStatistics.java
bgrozev: Added.

PR #56, opened 2016-03-27
Fix early DTLS

lyubomir:
The following modifications were applied while trying to remove a 1 second hard minimum on connectivity establishment between Videobridge and Chrome/WebRTC. Related modifications in jitsi-videobridge (i.e. which will be introduced in a branch there with the same name) (will) depend on these.

### DIFF
--- a/src/main/java/org/ice4j/ice/harvest/SinglePortUdpHarvester.java
+++ b/src/main/java/org/ice4j/ice/harvest/SinglePortUdpHarvester.java
@@ -85,10 +85,9 @@ public class SinglePortUdpHarvester
     public static List<SinglePortUdpHarvester>
         createHarvesters(int port)
     {
-        List<SinglePortUdpHarvester> harvesters
-           = new LinkedList<SinglePortUdpHarvester>();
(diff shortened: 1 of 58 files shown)


src/main/java/org/ice4j/ice/harvest/SinglePortUdpHarvester.java
```diff
@@ -303,21 +302,19 @@ private void runInHarvesterThread()
                     catch (SocketException se)
                     {
                         logger.info("Could not create a socket: " + se);
-                        continue;
```
bgrozev: This is not a critique, and I don't expect this addressed in any way. I'm just curious: Is there any specific reason these were removed? Syntactically, they don't do anything (or is there something subtle I'm missing?), but I think they give a hint to anyone reading the code about its intention.

lyubomir: I removed them because IntelliJ IDEA warned about them. I personally agreed with IntelliJ IDEA because the continues made me expect that there was more code bellow but it turned out there wasn't.

PR #58, opened 2016-04-02
Remove obsolete Ant-build and the embedded libraries

ibauersachs:
We use Maven now, no need to keep the embededd libraries and (probably outdated) ant-stuff around.


lyubomir:
Ah, @emcho may want to approve the removal of the Ant script because it contains run targets which we don't have in the Maven script. Otherwise, I'm OK to merge this PR.

ibauersachs:
The run-* targets? I'd say if anybody needs those, they are easily launched from an IDE without Ant. If they're really needed we can add them as Maven profiles using exec-maven-plugin.

lyubomir:
Yup, I agree, @ibauersachs. I wanted to make sure that @emcho is mentioned before we proceed with the merge.

emcho:
Having run scripts is an important part for ice4j development so, before merging this, could we please make sure we have alternatives in maven?

ibauersachs:
I've updated the PR and added the run targets.

PR #59, opened 2016-04-02
Use only OSS components for jain-sdp

ibauersachs:
The embedded jain-sdp library includes some interface-files that are
covered by a non-free license. Remove them and use a free version instead
that is published on Maven Central.


ibauersachs:
#59 and #60 depend on each other, creating a combined PR.

PR #62, opened 2016-04-13
Use an ArrayList instead of reimplementing it

champtar:
This should not change performance but is way more readable

Signed-off-by: Etienne CHAMPETIER <champetier.etienne@gmail.com>

### DIFF
--- a/src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
+++ b/src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
@@ -42,34 +42,6 @@ abstract class MultiplexingXXXSocketSupport
         = Logger.getLogger(MultiplexingXXXSocketSupport.class.getName());
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
```diff
@@ -259,29 +219,13 @@ else if (logger.isLoggable(Level.WARNING))
...
 
     /**
      * Initializes a new {@code MultiplexingXXXSocketSupport} instance.
-     *
-     * @param multiplexedXXXSocketClass the {@code Class} of the multiplexed
-     * sockets supported by the new instance
      */
-    protected MultiplexingXXXSocketSupport(
-            Class<MultiplexedXXXSocketT> multiplexedXXXSocketClass)
+    protected MultiplexingXXXSocketSupport()
```
ibauersachs: API break; please add a deprecated constructor taking the useless class type.

PR #68, opened 2016-04-18
ChannelData: fix padding

champtar:
Padding is mandatory for tcp but optionnal for udp
See https://tools.ietf.org/html/rfc5766#section-11.5

### DIFF
--- a/src/main/java/org/ice4j/message/ChannelData.java
+++ b/src/main/java/org/ice4j/message/ChannelData.java
@@ -188,15 +188,29 @@ public class ChannelData
      * @return a Message object constructed from the binMessage array
      * @throws StunException ILLEGAL_ARGUMENT if one or more of the arguments
      * have invalid values.
+     * @deprecated
      */
     public static ChannelData decode(byte binMessage[], char offset, char arrayLen) throws StunException
+    {
+        return decode(binMessage, offset);
+    }
+
+    /**
+     * Constructs a message from its binary representation.
+     * @param binMessage the binary array that contains the encoded message
+     * @param offset the index where the message starts.
+     * @return a Message object constructed from the binMessage array
+     * @throws StunException ILLEGAL_ARGUMENT if one or more of the arguments
+     * have invalid values.
+     */
+    public static ChannelData decode(byte binMessage[], char offset) throws StunException
     {
         char msgLen = 0;
         char channelNumber = 0;
         ChannelData channelData = null;
         byte data[] = null;
 
-        if(arrayLen < 4)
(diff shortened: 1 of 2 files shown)


ibauersachs:
Without looking at the actual padding change, I'm against merging this because it changes (removes) public API.

src/main/java/org/ice4j/message/ChannelData.java
```diff
@@ -196,7 +196,12 @@ public static ChannelData decode(byte binMessage[], char offset, char arrayLen)
         }
 
         msgLen = (char)((binMessage[offset++]<<8) | (binMessage[offset++]&0xFF));
-        if(msgLen != (binMessage.length - 4))
+        /*
+         * message can be padded
+         * https://tools.ietf.org/html/rfc5766#section-11.5
+         */
+        if (msgLen != (binMessage.length - HEADER_LENGTH)
+            && ((msgLen % 4) + 4) != (binMessage.length - HEADER_LENGTH))
```
ibauersachs: I might be tired and don't know ICE or TURN, but I'd assume that `binMessage.length - HEADER_LENGTH` can be legally greater than `[4,7]`.

champtar: yep this is wrong, you should read ```msgLen - (msgLen % 4) + 4```

champtar: or even ```((msgLen - 1) / 4 + 1) * 4``` which is better
8/9/10/11 gives 8/12/12/12

champtar:
I've reworked this PR to not break current API

ibauersachs:
Apart from your good fix with the offset in decode, I don't quite see the use case for the other changes. Is there a problem if we always pad the messages?
Also:
- getDataLength() method still behaves differently (probably doesn't really matter)
- padTo4 as a separate method is overkill and pollutes the API, at the very least make it private

I'm still reluctant to change this at all. Can someone else please give some input? @bgrozev?

champtar:
hi @ibauersachs,
We can always pad according to the RFC, but i prefer to be sure to keep back compat when possible.
getDataLength() is unchanged (it behaves exactly the same as before)
padTo4 is a bit unreadable, thus the separate function, but i will make it private.

I started this PR because length in getDataLength() was a dead store, and FindBugs spotted it.

ibauersachs:
Oh, right, it was too late yesterday and I missed that getDataLength was returning the array length instead of the calculated value. So, okay, after making padto4 private I'm okay merging it.

champtar:
padTo4 now private

PR #73, opened 2016-06-01
handover IOException to higher level is needed

ubivent-seuffert:
the DelegatingDatagramSocket needs to hand a send IOException to a higher level, otherwise the socket can not be closed on send failures


ubivent-seuffert:
CLA signed.

ibauersachs:
Hmm. Is that dirty hack still necessary? Last I've read is that OSX now has a working IPv6 stack.

bgrozev:
Is there any reason to not simplify this to match for NoRouteToHostException in the catch clause? We're still swallowing exceptions that are not IOException.

ibauersachs:
I doubt that this check is still necessary at all. But I don't have a Mac (and neither an IPv6 network) to test.

PR #77, opened 2016-08-11
Exposes RateStatistics#getAccumulatedCount

bgrozev:
and synchronizes access to some of the methods.

### DIFF
--- a/src/main/java/org/ice4j/util/RateStatistics.java
+++ b/src/main/java/org/ice4j/util/RateStatistics.java
@@ -74,7 +74,7 @@ public class RateStatistics
         this.scale = scale / (buckets.length - 1);
     }
 
-    private void eraseOld(long nowMs)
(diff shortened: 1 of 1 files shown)


gpolitis:
jenkins it's ok to test

src/main/java/org/ice4j/util/RateStatistics.java
```diff
@@ -102,25 +102,45 @@ private void eraseOld(long nowMs)
         oldestTime = newOldestTime;
     }
 
+    public long getRate()
```
gpolitis: I'm not sure why we don't have any JavaDoc for (at least) the public methods of this class, but I don't see a reason why we should continue this practice.

gpolitis:
Could you please add yourself to the authors of this class?

gpolitis:
This class appears to be thread safe now. Could you please mention that in the class-level JavaDoc?

PR #83, opened 2016-09-27
Do not use String.format and construct the string only if logger…

matteocampana:
…ST is enabled.


matteocampana:
CLA already signed for other Jitsi projects.

bgrozev:
jenkins add to whitelist

bgrozev:
Thank you @matteocampana!

PR #87, opened 2016-11-17
Harvesting

bgrozev:
Moves the mapping harvesters' configuration from jitsi-videobridge to ice4j. Automatically adds mapping harvesters to Agents. Re-implements the STUN mapping harvester in a way which allows discrimination between a failure of the STUN server and a match between the local and public addresses.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -1,0 +1,79 @@
+# General
+This file describes some of the java properties which ```ice4j``` uses
+to configure itself.
+
+## Interfaces and IP addresses
+
+### *org.ice4j.ice.harvest.ALLOWED_INTERFACES*
+Default: all interfaces are allowed.
+
+This property can be used to specify a ";"-separated list of interfaces which are
+allowed to be used for candidate allocations. If not specified, all interfaces are
+considered allowed, unless they are explicitly blocked (see below).
+
+### *org.ice4j.ice.harvest.BLOCKED_INTERFACES*
+Default: no interfaces are blocked.
+
+This property can be used to specify a ";"-separated list of interfaces which are
+not allowed to be used for candidate allocations. 
+
+### *org.ice4j.ice.harvest.ALLOWED_ADDRESSES*
+Default: all addresses are allowed.
+
+This property can be used to specify a ";"-separated list of IP addresses which
+are allowed to be used for candidate allocations. If not specified, all addresses
+are considered allowed, unless they are explicitly blocked (see below).
+
+### *org.ice4j.ice.harvest.BLOCKED_ADDRESSES*
+Default: no addresses are blocked.
+
+This property can be used to specify a ";"-separated list of IP addresses which
+are not allowed to be used for candidate allocations. 
+
+### *org.ice4j.ipv6.DISABLED*
+Type: boolean
+
+Default: false
+
+This property can be used to disable binding on IPv6 addresses.
+
+
+## Mapping harvesters
+Ice4j uses the concept of "mapping harvesters" to handle known IP address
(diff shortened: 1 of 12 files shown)


ibauersachs:
Is it possible to keep the existing StunMappingCandidateHarvester constructors so that the public API stays intact?

bgrozev:
Thanks for the comment Ingo!

> Is it possible to keep the existing StunMappingCandidateHarvester constructors so that the public API stays intact?

Adding dummy constructors to keep the API would be easy, but that would not keep the previous functionality. I tend to prefer breaking the API since fixing any applications using it will be quite easy -- just stop using StunMappingCandidateHarvester (I don't see any reason to use it from outside ice4j, actually, might as well make it protected).

But I don't mind adding dummy constructors -- let me know if you prefer to do it this way.

ibauersachs:
I did a comparison and the API is already broken, mostly because of #81. See the [diff](https://gist.githack.com/ibauersachs/750fb082f536ed17b7b72a092f8f4c09/raw/6b07babfd35f45dc5f366a946510c89778557237/ice4j-1.0%20vs%20ice4j-1.1-snapshot.html) of the current master (vs. [diff of this pr](https://gist.githack.com/ibauersachs/0a310c3055f045ef7046978c8ab42d18/raw/29b84f6832aa193b0497a5dde1c8fc4cdf80365c/ice4j-1.0%20vs%20ice4j-1.1-snapshot-pr87.html)). Can you thus please update the version to 2.0-SNAPSHOT?

And what is the state of ice4j in general? Is there a lot more to come, can a 2.0 be released with now, soon, or ...?

bgrozev:
> Can you thus please update the version to 2.0-SNAPSHOT?

Sure.

> And what is the state of ice4j in general? Is there a lot more to come, can a 2.0 be released with now, soon, or ...?

There is more to come in the next months, but I don't know API changes will be necessary. This wasn't an important consideration for me for these two PRs. About the release -- I don't know, because I don't understand how releases are being used.

gpolitis:
I would expect some major API changes if we were to bump the version to 2.0. This is "just" an improvement of the existing code that does indeed break the 1.1 API, but 1.1 hasn't been released yet, so small changes in the API should be expected.

mondain:
I merged this with my fork and it seems a bit faster than older versions; awesome work!

ibauersachs:
@bgrozev We have 1.0 published on Maven Central and I know that some other projects (e.g. from Daniel Pocock) are using it.

@gpolitis True, but: changes in the API for 1.1 would be perfectly fine as long as they're additions, not removals. Right now a build of the master branch (with or without this PR) could not be an in-place binary replacement.

But maybe it would be necessary to redefine what is public API and what is not, document that, and then accept breakage in certain classes even though they're public.

bgrozev:
@ibauersachs Thanks for the explanation! So you think that 1.1, when released, should be able to replace the 1.0 binary in-place?

If that's the case I suggest that we:
1. Do a 1.1 release with the latest version which doesn't break if the binary is replaced
2. Move to 2.0 and accept that it will not be an in-place binary replacement for 1.0 or 1.1.

Does that make sense?

bgrozev:
@mondain this working in any way faster than master would be an unknown, but welcome, side effect :)

ibauersachs:
@bgrozev Releasing a non-breaking 1.1 would be something from >2 months ago. Not sure if that makes sense anymore, but we can do that. Breaking changes in 2.0 are perfectly okay, see http://semver.org/.

bgrozev:
OK, moving on with 2.0. Thanks!

Jenkins: ok to test

PR #88, opened 2016-11-21
Virtual socket layer

bgrozev:
Introduces a "virtual socket layer" in the form of a MergingDatagramSocket class, and an instance for each Component. The feature is optional and OFF by default.

### DIFF
--- a/src/main/java/org/ice4j/ice/ComponentSocket.java
+++ b/src/main/java/org/ice4j/ice/ComponentSocket.java
@@ -17,6 +17,7 @@
  */
 package org.ice4j.ice;
 
+import org.ice4j.*;
 import org.ice4j.socket.*;
 
 import java.beans.*;
@@ -58,6 +59,12 @@ public class ComponentSocket
      * The owning {@link Component}.
      */
     private Component component;
+
+    /**
+     * Whether we have invoked {@link #initializeActive}.
+     */
+    private boolean initializedActive = false;
+
 
     /**
      * Initializes a new {@link MergingDatagramSocket} instance.
@@ -129,8 +136,7 @@ public class ComponentSocket
             return;
         }
 
-        if (IceMediaStream.PROPERTY_PAIR_STATE_CHANGED.equals(
-            propertyName))
+        if (IceMediaStream.PROPERTY_PAIR_STATE_CHANGED.equals(propertyName))
         {
             CandidatePairState newState
                 = (CandidatePairState) event.getNewValue();
@@ -140,6 +146,36 @@ public class ComponentSocket
                 addAuthorizedAddress(
                         pair.getRemoteCandidate().getTransportAddress());
             }
+        }
+        else if (IceMediaStream.PROPERTY_PAIR_NOMINATED.equals(propertyName))
+        {
+            if (initializedActive)
+            {
+                return;
+            }
+            initializedActive = true;
+
+            // Find the remote address and the correct socket to be used by
+            // the pair.
+            LocalCandidate localCandidate = pair.getLocalCandidate();
+            LocalCandidate base = localCandidate.getBase();
+            if (base != null)
+                localCandidate = base;
+
+            TransportAddress remoteAddress = null;
+            RemoteCandidate remoteCandidate = pair.getRemoteCandidate();
+            if (remoteCandidate != null)
+            {
+                remoteAddress = remoteCandidate.getTransportAddress();
+            }
+
+            // The local candidate may have more than one associated socket.
+            // Make sure we get the one for the remote address that we are
+            // going to use.
+            IceSocketWrapper socketWrapper
+                = localCandidate.getCandidateIceSocketWrapper(remoteAddress);
+
+            initializeActive(socketWrapper, remoteAddress);
         }
     }
 
--- a/src/main/java/org/ice4j/ice/ComponentSocket.java
+++ b/src/main/java/org/ice4j/ice/ComponentSocket.java
@@ -1,0 +1,161 @@
+/*
+ * ice4j, the OpenSource Java Solution for NAT and Firewall Traversal.
+ *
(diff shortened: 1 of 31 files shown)


src/main/java/org/ice4j/ice/ComponentSocket.java
```diff
@@ -0,0 +1,206 @@
...
+    private final Object authorizedAddressesSyncRoot = new Object();
+
+    /**
+     * The set of remote addresses, which this socket is allowed to receive
+     * from. These should be the addresses which we have confirmed (e.g. by
+     * having received a STUN message with correct authentication fields).
+     */
+    private Set<SocketAddress> authorizedAddresses
+        = Collections.newSetFromMap(
+                new ConcurrentHashMap<SocketAddress, Boolean>());
```
bbaldino: i'm a bit confused by this...why back this set with a concurrent hashmap?  also, why back it with a map at all?  we create the map in place here, so there's no accessing it directly.  why not just create a set?  if the purpose of backing it by the concurrenthashmap is to inherit the thread safety, it seems like we don't get to take full advantage of that anyway as half the accesses are in a synchronized block (although i'm sure 'accept' is called much more frequently, so maybe this is all just to optimize for that path?)

src/main/java/org/ice4j/ice/ComponentSocket.java
```diff
@@ -0,0 +1,206 @@
...
+                return;
+            }
+
+            logger.info("Adding allowed address: " + address);
+
+            Set<SocketAddress> newSet
+                = Collections.newSetFromMap(
+                        new ConcurrentHashMap<SocketAddress, Boolean>());
+            newSet.addAll(authorizedAddresses);
+            newSet.add(address);
```
bbaldino: probably a java thing i'm missing...but why can't we just add the new address to the set directly?  it seems like that wouldn't break the 'accept' method, but maybe i'm missing something.

src/main/java/org/ice4j/ice/ComponentSocket.java
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * {@inheritDoc}
+     * </p>
+     * Handles property change events coming from ICE pairs.
+     * @param event
+     */
+    @Override
+    public void propertyChange(PropertyChangeEvent event)
+    {
+        String propertyName = event.getPropertyName();
```
bbaldino: minor nit but technically this could lower (right above where it's used) so we don't have to allocate it if we're not gonna use it anyway

src/main/java/org/ice4j/ice/ComponentSocket.java
```diff
@@ -0,0 +1,206 @@
...
+                        pair.getRemoteCandidate().getTransportAddress());
+            }
+        }
+        else if (IceMediaStream.PROPERTY_PAIR_NOMINATED.equals(propertyName))
+        {
+            if (initializedActive)
+            {
+                return;
+            }
+            initializedActive = true;
```
bbaldino: minor nit: maybe we should put this right above the actual call to initalizeActive?  that way it's less likely some return/error case could occur/be inserted between the two and end up in a bad state?

src/main/java/org/ice4j/ice/DefaultNominator.java
```diff
@@ -314,18 +315,24 @@ public RelayedCandidateTask(CandidatePair pair)
...
-             * the relayed candidate
-             */
+            // Make it clear that PROPERTY_CHECK_LIST_CHECKS is in use here.
+            if (!CheckList.PROPERTY_CHECK_LIST_CHECKS.equals(
+                        evt.getPropertyName()))
+            {
+                return;
+            }
+
+            // check list has running out of ordinary checks, see if all other
```
bbaldino: i know you just copy-pasted this, but looks like there is a typo?  is it "check list has **run** out of ordinary checks"?

src/main/java/org/ice4j/ice/ComponentSocket.java
bgrozev: Good point, will fix.

src/main/java/org/ice4j/ice/ComponentSocket.java
bgrozev: I moved it after the call to initializeActive, and added a synchronized block (will push in a second).

bgrozev: Do you think updating the boolean after the call to initializeActive is OK?

src/main/java/org/ice4j/ice/DefaultNominator.java
bgrozev: Fixed.

src/main/java/org/ice4j/ice/ComponentSocket.java
bbaldino: it seems like it's being used as a way to prevent it from being called twice, so maybe need to stay before the call itself?

bgrozev: The "if" part stays at the top, the initializedActive=true part is moved after the actual call (see 7cac0e50a7fb55ca0e7fb2ae7d65a17cbff85570)

bbaldino: ok, so is it true that only a single thread will go through that code?  otherwise a second thread could enter and make a second call to initializeActive if the first one didn't finish and set the flag yet. this is making me realize that we probably need to either: a) move the if down as well, but then we've potentially allocated some stuff/did some work we're not going to use or b) move the setting of the flag back up.  i just worry with it being up that it's more prone to having an early exit or exception inserted that results in initializeActive never getting called.  maybe that's too paranoid though and it's fine the way it was. :/

bgrozev: Forgot to mention: also added a synchronized block.

PR #89, opened 2016-12-13
Tcp fallback

bgrozev:


src/main/java/org/ice4j/ice/Component.java
```diff
@@ -902,6 +922,12 @@ public RemoteCandidate findRemoteCandidate(TransportAddress remoteAddress)
      */
     protected void setSelectedPair(CandidatePair pair)
     {
+        if (keepAliveStrategy == KeepAliveStrategy.SELECTED_ONLY)
```
bbaldino: since the keep-alive strategy can be changed at any time, i'm wondering if maybe it makes sense to keep track of all pairs and then just denote one of them as 'selected'?  that way we don't lose track of any of them.  might be a rare thing that the strategy is changed mid-way through, but does look like it's technically possible?

bbaldino: or, if we don't want to let it be changed, maybe we should just pass the keep-alive strategy in the ctor rather than having a setter?

bgrozev: Just saw your second message, I'll do that.

bgrozev: I check the state of the Agent instead, and throw an exception if it's been started. Would have had to add parameters in 3 places otherwise.

src/main/java/org/ice4j/ice/Component.java
```diff
@@ -1079,11 +1087,20 @@ else if (keepAliveStrategy == KeepAliveStrategy.SELECTED_AND_TCP)
...
     /**
      * Sets the {@link KeepAliveStrategy} for this {@link Component}.
-     * @param strategy the strategy to set.
+     * @param strategy the strategy to set. The keep-alive strategy can only
+     * be modified before the parent {@link Agent} is started.
+     * @throws IllegalStateException if the agent has been started.
      */
     public void setKeepAliveStrategy(KeepAliveStrategy strategy)
     {
+        IceProcessingState agentState
```
bbaldino: any reason to keep the setter and throw if it's too late as opposed to removing the setter and just passing the keepalive strategy in the constructor?  that would achieve the goal and prevent any chance of the caller making a mistake that would cause an exception.

bbaldino: oops...just saw your other comment as to why @bgrozev.  i still think doing it in the constructor conveys the intention better. if you added a constructor which didn't take the parameter but defaulted it to the current default, would that prevent having to change the other places?

PR #90, opened 2016-12-21
fix: Fixes a failure to recognize consent

bgrozev:
confirmation for TCP pairs.

### DIFF
--- a/src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
+++ b/src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
@@ -739,8 +739,11 @@ class ConnectivityCheckClient
         }
 
         // Selected pairs get their consent freshness confirmed.
-        if (checkedPair == checkedPair.getParentComponent().getSelectedPair())
+        // XXX Should we also confirm consent freshness for non-selected pairs?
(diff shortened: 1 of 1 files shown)


src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
```diff
@@ -739,8 +739,11 @@ else if (!parentAgent.isControlling()
         }
 
         // Selected pairs get their consent freshness confirmed.
-        if (checkedPair == checkedPair.getParentComponent().getSelectedPair())
+        // XXX Should we also confirm consent freshness for non-selected pairs?
+        if (checkedPair.equals(checkedPair.getParentComponent().getSelectedPair()))
```
bbaldino: nice catch.  your comment sounds right...if we expect to be able to fall back, do we need to be carrying this out for the non-selected pairs as well?

bgrozev: Probably, yeah. But I haven't looked at potential side effects.

bbaldino: ok, well lgtm for now

PR #91, opened 2016-12-22
Make sure the new messageProcess is added back

kgdev:


mondain:
This looks like an oversight; bumping for progress.

PR #92, opened 2017-01-07
Networkutils rfc2732

ibauersachs:
Previously bracketed addresses provided to strToIPv6 would return null.


bgrozev:
jenkins ok to test

bgrozev:
That's a great example of an appropriate thing to unit-test.

PR #96, opened 2017-02-24
Allows some form of mobility

bgrozev:
Allows new sockets and pairs to be added to an ICE Agent which has already reached the Completed or Terminated state. The new socket will be added to the Component's ComponentSocket, and if the pair succeeds the new remote address will be allowed.

### DIFF
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -1541,19 +1541,8 @@ public class Agent
                                          String           localUFrag,
                                          boolean          useCandidate)
     {
-        if (isOver())
-        {
-            //means we already completed ICE and are happily running media
-            //the only reason we are called is most probably because the remote
-            //party is still sending Binding Requests our way and making sure we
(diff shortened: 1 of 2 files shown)


src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -1541,19 +1541,8 @@ protected void incomingCheckReceived(TransportAddress remoteAddress,
                                          String           localUFrag,
                                          boolean          useCandidate)
     {
-        if (isOver())
```
gpolitis: Not really important so please feel free to ignore, but I was wondering if we could remove the if statement but keep and amend the comment?

bgrozev: I removed the comment, because I don't think it's accurate anymore. It could be a BR on the selected pair, or a PR for another succeeded pair, or a completely new pair.

bgrozev: s/PR/Binding Request/

PR #104, opened 2017-03-27
	change MultiplexingXXXSocketSuppport::receive to 'trade' buffers in order to avoid a packet buffer allocation/copy.

bbaldino:
Currently, even though the passed-in packet has a buffer already allocated, we allocate a new buffer to receive a packet and put it on the buffer.  When we're ready to fill the given packet with the received data, we copy into the given buffer.
Instead, we can 'trade' the buffers: the passed-in packet's buffer will be used to perform the receive and then added to the queue.  The packet at the front of the queue (which will be discarded anyway) will have its buffer 'stolen' by the passed-in
packet instance. By doing this we can avoid a large buffer allocation and a copy for every packet.

### DIFF
--- a/src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
+++ b/src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
@@ -140,9 +140,12 @@ abstract class MultiplexingXXXSocketSupport
      */
     public static void move(DatagramPacket src, DatagramPacket dest)
     {
-        dest.setAddress(src.getAddress());
-        dest.setPort(src.getPort());
-        dest.setData(src.getData());
+        synchronized (dest)
+        {
+            dest.setAddress(src.getAddress());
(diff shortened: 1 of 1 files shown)


bbaldino:
@lyubomir could you take a look at this?  maybe there's a better way to do this in java (stealing the buffer) than what i have here...i'm not crazy about the fact that there's no good way in `move` to set the `src` back to an uninitialized state (the `DatagramPacket` setters don't allow it).  i've tested it and it's a nice win, though.  the place where we were doing this allocation accounted for roughly 5% of all of our `byte[]` allocations, and in addition to saving that, we also save the copy.

bbaldino:
also, i'm still looking into this a bit more and make sure it hasn't broken anything, so just interested in any feedback at this point (not merging yet).

bbaldino:
unfortunately this isn't as big of a win...i just found a bug where i had to had another copy to keep things happy, so we don't gain much from this change alone.  closing.

PR #111, opened 2017-04-14
Marshalling byte[] based on JDK class.

MarcinMiotk:


MarcinMiotk:
I've already signed CLA.

pom.xml
```diff
@@ -8,7 +8,7 @@
   </parent>
 
   <artifactId>ice4j</artifactId>
-  <version>2.0-SNAPSHOT</version>
+  <version>2.0-mami-SNAPSHOT</version>
```
ubivent-seuffert: accidentally renamed?

PR #113, opened 2017-04-28
Introduce harvesting timeout option

SStorm:
I've been seeing that occasionally one harvester goes into a state where it never finishes, thus blocking the the entire execution. I've seen this happen when running inside of a Docker container with host networking (--net=host) where ice4j has access to Docker's vethXYZ interfaces. We then have it stuck like this:

```
   java.lang.Thread.State: WAITING
        at sun.misc.Unsafe.park(Native Method)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:429)
        at java.util.concurrent.FutureTask.get(FutureTask.java:191)
        at org.ice4j.ice.harvest.CandidateHarvesterSet.harvest(CandidateHarvesterSet.java:200)
        at org.ice4j.ice.harvest.CandidateHarvesterSet.harvest(CandidateHarvesterSet.java:118)
        at org.ice4j.ice.harvest.CandidateHarvesterSet.harvest(CandidateHarvesterSet.java:98)
        at org.ice4j.ice.Agent.gatherCandidates(Agent.java:604)
        at org.ice4j.ice.Agent.createComponent(Agent.java:508)
        at org.ice4j.ice.Agent.createComponent(Agent.java:455)
```

While the harvester itself is doing:
```
   java.lang.Thread.State: WAITING
        at java.lang.Object.wait(Native Method)
        at java.lang.Object.wait(Object.java:502)
        at org.ice4j.ice.harvest.StunCandidateHarvester.waitForResolutionEnd(StunCandidateHarvester.java:398)
        at org.ice4j.ice.harvest.StunCandidateHarvester.harvest(StunCandidateHarvester.java:256)
        at org.ice4j.ice.harvest.CandidateHarvesterSetElement.harvest(CandidateHarvesterSetElement.java:86)
        at org.ice4j.ice.harvest.CandidateHarvesterSetTask.run(CandidateHarvesterSetTask.java:107)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

The timeout could also be added to the StunCandidateHarvester, but having it on all harvesters seemed like a better idea. Happy to modify to fit any coding conventions and the like.

### DIFF
--- a/src/main/java/org/ice4j/StackProperties.java
+++ b/src/main/java/org/ice4j/StackProperties.java
@@ -233,6 +233,12 @@ public class StackProperties
             = "org.ice4j.ice.harvest.USE_DYNAMIC_HOST_HARVESTER";
 
     /**
+     * Timeout, in seconds, of how long to wait for an individual harvest before timing out
+     */
+    public static final String HARVESTING_TIMEOUT
+            = "org.ice4j.ice.harvest.HARVESTING_TIMEOUT";
+
+    /**
      * Returns the String value of the specified property (minus all
      * encompassing whitespaces)and null in case no property value was mapped
      * against the specified propertyName, or in case the returned property
--- a/src/main/java/org/ice4j/ice/harvest/CandidateHarvesterSet.java
+++ b/src/main/java/org/ice4j/ice/harvest/CandidateHarvesterSet.java
@@ -21,6 +21,7 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.logging.*;
 
+import org.ice4j.*;
 import org.ice4j.ice.*;
 
 /**
@@ -197,7 +198,17 @@ public class CandidateHarvesterSet
             {
                 try
                 {
-                    future.get();
+                    future.get(StackProperties.getInt(StackProperties.HARVESTING_TIMEOUT, 15), TimeUnit.SECONDS);
+                    break;
+                }
+                catch (TimeoutException te)
+                {
+                    CandidateHarvesterSetElement harvester = task.getKey().getHarvester();
+                    if (harvester != null)
+                    {
+                        harvester.setEnabled(false);
+                    }
+                    logger.warning("timed out while harvesting from " + harvester);
                     break;
                 }
                 catch (CancellationException ce)


bgrozev:
Thank you for the patch! I think the functionality is OK to merge, just a couple of small things:

-- Why does it not break out of the loop on TimeoutException (but it does break out on other exception)?
-- We'll need the CLA signed
-- Convention: please keep the importa package only (e.g. use ".*", not ".Component"). Please add braces on all ifs, even one liners.

Jenkins, please test

SStorm:
1. Should it? Only one harvester has timed out, it's not a fatal exception. We should presumably still wait for the others to finish? Tell me if that's not the case and I'll change
2. Company legal currently looking into that, hoping to sign shortly
3. Fixed

bgrozev:
> Should it? Only one harvester has timed out, it's not a fatal exception. We should presumably still wait for the others to finish? Tell me if that's not the case and I'll change

It seems to me that we should break out of the inner loop, otherwise we'll be calling future.get() again for the same harvester. This would still leave us looping over the rest of the harvesters. Or am I missing something?

I also wonder if we should setEnabled(false) as in the case of an ExecutionException.

SStorm:
Yes you're right, didn't notice the loop there.

bgrozev:
Thank you! Looks good to me now. I'll merge this, but it may be a few days before we get the CLA on our end (the process still requires human work, and it's from another team).

Jenkins, test this please

SStorm:
CLA signed, thanks.

SStorm:
Ping? :)

bgrozev:
Sorry, we just got the PR 20 minutes ago. Thanks!

PR #114, opened 2017-05-10
specify where to apply configurations

pedro-nonfree:
According to https://github.com/jitsi/jitsi-meet/issues/1390#issuecomment-300355464 configurations of `ice4j` are applied in the same place as `jitsi-videobridge`.

At the moment this is not clear in ice4j/doc/configuration.md

As a proposal to avoid (this/my) confusion I used a piece of text from [jitsi-videobridge/doc/http.md](https://github.com/jitsi/jitsi-videobridge/blob/9f13b7365512b946eff1703838b496aeacebda25/doc/http.md)

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -1,6 +1,7 @@
 # General
-This file describes some of the java properties which ```ice4j``` uses
-to configure itself.
+This file describes some of the java properties which ```ice4j``` uses to configure
+itself in /etc/jitsi/videobridge/sip-communicator.properties if installed from package,
+or in $HOME/.sip-communicator/sip-communicator.properties if running from source.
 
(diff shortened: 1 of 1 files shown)


damencho:
I don't think the text is in the correct place. There should be a text in the videobridge itself, something like: further configuration can be made using ice4j props and link to this page. Ice4j is standalone lib and is used in many places different than jvb.

pedro-nonfree:
> Ice4j is standalone lib and is used in many places different than jvb.

I didn't think about that. Hence, I agree with your solution. How we proceed? I suggest you changing the text the way you are saying. Alternatively I can propose a pull request in jitsi-videobridge.

damencho:
I will do it. Thanks for the proposal.

PR #119, opened 2017-06-23
fix: Works around a bug in Edge

bgrozev:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12332457/
https://github.com/jitsi/lib-jitsi-meet/issues/498

### DIFF
--- a/src/main/java/org/ice4j/attribute/UsernameAttribute.java
+++ b/src/main/java/org/ice4j/attribute/UsernameAttribute.java
@@ -17,6 +17,8 @@
  */
 package org.ice4j.attribute;
 
+import org.ice4j.*;
+
 import java.util.*;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/ice4j/attribute/UsernameAttribute.java
```diff
@@ -56,8 +76,17 @@
      *   offset is equal to the index of the first byte after length)
      * @param length the length of the binary array.
      */
+    @Override
     void decodeAttributeBody(byte[] attributeValue, char offset, char length)
     {
+        if (stripTrailingZeroes)
+        {
+            while (length > 0 && attributeValue[offset + length] == 0)
```
bbaldino: a nitpick i guess, but given there's no minimum length, i'm wondering if we should prevent this from going to 0 length?  it seems arbitrary to stop it at length 1.

bbaldino: ignore these mad ramblings.

PR #121, opened 2017-07-11
Multiplexing fixes and features

bgrozev:


src/main/java/org/ice4j/socket/MultiplexingDatagramSocket.java
```diff
@@ -274,6 +384,26 @@ public MultiplexedDatagramSocket getSocket(
...
 
     /**
+     * Adds a specific {@link MultiplexedDatagramSocket} to this instance. The
+     * given socket must be already associated with this instance.
+     * @param multiplexed the socket to add.
+     * @throws IllegalArgumentException if the given socket is not associated
+     * with this instance, or its filter is already used for another multiplexed
+     * socket.
+     */
+    public void addSocket(MultiplexedDatagramSocket multiplexed)
```
bbaldino: i'm confused by the need for this...if it must already be associated with this, why do we allow it to become disassociated?

src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
```diff
@@ -388,6 +391,38 @@ public MultiplexedXXXSocketT getSocket(DatagramPacketFilter filter)
...
     /**
+     * Adds a specific {@link MultiplexedXXXSocketT} instance to the list of
+     * filtered sockets of this instance. The filter which is to determine
+     * which packets are accepted by the socket is the socket's existing
+     * filter obtained via {@link MultiplexedXXXSocketT#getFilter()}.
+     * @param multiplexed the socket instance to add.
+     * @throws IllegalArgumentException if there already exists a socket
+     * with the specified socket's filter.
+     */
+    void addSocket(MultiplexedXXXSocketT multiplexed)
```
bbaldino: same as above

src/main/java/org/ice4j/socket/MultiplexingDatagramSocket.java
bgrozev: Because I want to override the receive() method of the MultiplexedDatagramSocket from code in the bridge. I guess I could expose a callback somehow, but this was the easiest solution I found.

PR #122, opened 2017-07-12
Multiplexing fixes and features

bgrozev:
Removes the extra API in Multiplexing/Multiplexed DatagramSocket as discussed offline with @bbaldino .

### DIFF
--- a/src/main/java/org/ice4j/socket/MultiplexingDatagramSocket.java
+++ b/src/main/java/org/ice4j/socket/MultiplexingDatagramSocket.java
@@ -99,8 +99,9 @@ public class MultiplexingDatagramSocket
 
     /**
      * The list of <tt>DatagramPacket</tt>s to be received through this
-     * <tt>DatagramSocket</tt> i.e. not accepted by the <tt>DatagramFilter</tt>s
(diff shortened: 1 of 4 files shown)


src/main/java/org/ice4j/socket/MultiplexingDatagramSocket.java
```diff
@@ -138,6 +148,25 @@ public int getReceiveBufferSize()
...
+     * @throws SocketException if the socket could not be opened, or the socket
+     * could not bind to the specified local port
+     * @param persistent whether this socket should be kept open after all of
+     * its {@link MultiplexedDatagramSocket}s are closed.
+     * @see DatagramSocket#DatagramSocket()
+     */
+    public MultiplexingDatagramSocket(boolean persistent)
+        throws SocketException
+    {
+        this.persistent = persistent;
```
bbaldino: should this do `this(null, false)` and there be a null check for the delegate in the other ctor maybe? just an idea.

bbaldino: ok now i see there are all sorts of ctors here :/ i was hoping we could keep the main ctor logic to one of them but doesn't look like this would help much.

Issue #95, opened 2017-01-17
New release?


SStorm:
Any plans on making a new release? Latest available in maven central is 1.0 which is very old - if you're relying on any new features / bugfixes since then (like me), you need to build and upload a version yourself, which is quite cumbersome to say the least...

ibauersachs:
I didn't upload 1.1 because it accidentally introduced breaking API changes. So does 2.0 and AFAIK there's currently quite some development going on for it.
@bgrozev Any comment on when there might be a stable state that could go as 2.0 - and keeps the public API for the foreseeable future?
@damencho Can we upload the -SNAPSHOT builds to Maven Central too? (We have an account for Jenkins prepared that can upload snapshots only to org.jitsi)

OrangeDog:
Could upload 1.1 as 2.0, and the current 2.0 can be 3.0?

ibauersachs:
Current 2.0 was never released as such. Before I do a next release however, I want to have a plan for further releases.

ibauersachs:
Proposal for future stable releases:

1. Switch to sematic versioning
2. Merge https://github.com/jitsi/ice4j/issues/95 to enforce 1.
3. Deploy snapshots to Maven Central (we have a Jenkins user for this)
4. Release whatever version used in a Jitsi Meet stable-combo-release also to Maven Central

@jitsi/developers please vote this comment with thumbs up/down (and leave comments).

OrangeDog:
I assume you mean merge #123, not #95.

ibauersachs:
@OrangeDog yes, edited, thanks!

lecogiteur:
Hi,

I don't see new release from 2017. This project seems alive. Do you release it? When?

bgrozev:
We just use the snapshots in our projects and haven't bothered with making a release.

lecogiteur:
@bgrozev how to manage breaking change of library, we need to fork it?

OrangeDog:
@bgrozev that is a foolish practice for your own projects, and very unhelpful for an open-source library.
While people can [lock their snapshots](https://www.mojohaus.org/versions-maven-plugin/examples/lock-snapshots.html), it still places a huge burden working out what's changed and which snapshots are safe to use.

ibauersachs:
[...]

PR #123, opened 2017-07-12
Preparations for releasing to Maven Central

ibauersachs:
This change to the POM will fail the build when the project version doesn't adhere to semantic versioning, i.e. when the source or binary compatibility is broken a new major version must be set.

This is a preparation for #95.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
   </parent>
 
   <artifactId>ice4j</artifactId>
-  <version>2.0-SNAPSHOT</version>
+  <version>2.0.0-SNAPSHOT</version>
   <packaging>bundle</packaging>
 
   <name>ice4j</name>
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,38 @@
           </includes>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>com.github.siom79.japicmp</groupId>
+        <artifactId>japicmp-maven-plugin</artifactId>
+        <version>0.10.0</version>
+        <configuration>
+          <newVersion>
+            <file>
+              <path>${project.build.directory}/${project.build.finalName}.jar</path>
+            </file>
+          </newVersion>
+          <oldVersion>
+            <dependency>
+              <groupId>${project.groupId}</groupId>
+              <artifactId>${project.artifactId}</artifactId>
+              <version>1.0.0</version>
+            </dependency>
+          </oldVersion>
+          <parameter>
+            <onlyModified>true</onlyModified>
+            <breakBuildBasedOnSemanticVersioning>true</breakBuildBasedOnSemanticVersioning>
+            <reportOnlyFilename>true</reportOnlyFilename>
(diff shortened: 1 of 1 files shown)


bgrozev:
FWIT this looks good to me. But I'm trying to test it by removing a public method and running "mvn install" and I don't see any failures. Is this expected?

ibauersachs:
@bgrozev Yes. The plugin compares 2.0.0-SNAPSHOT against 1.0.0. Since the major version was increased, the entire API can change. If you want to play around, try changing the version in master to e.g. 1.1.0-SNAPSHOT and you should see it fail.

gpolitis:
The ice4j version in the JVB pom file needs to be adjusted and pushed to a branch called semver:

    00:08:21 [ERROR] /home/jenkins/jitsi-videobridge/src/main/java/org/jitsi/videobridge/osgi/JvbBundleConfig.java:[250,39] cannot find symbol
    00:08:21   symbol:   variable STRIP_TRAILING_ZEROES_PNAME
    00:08:21   location: class org.ice4j.attribute.UsernameAttribute

damencho:
@gpolitis if you are talking about why the tests fail, this will not work. As pr testing replaces ice4j version with 2.0.0-snapshot everywhere ice4j is used when it detects a change in it. This is the reason tests had failed. But if I change it, it will fail for other tests :)

gpolitis:
oh, I see, thanks for explaining @damencho. We merge without the tests then?

ibauersachs:
@damencho I don't quite understand what you said. If the PR testing replaces the version of ice4j in all dependent projects with the version of this PR branch, then why did it fail? The STRIP_TRAILING_ZEROES_PNAME field is clearly present here.

damencho:
It's that we are replacing with 2.0-SNAPSHOT where this one is 2.0.0-SNAPSHOT, I'm really not sure what maven gets to compile it in this case. I can try finding a free spot today without tests and change it just for to run this one.

ibauersachs:
So where does 2.0-SNAPSHOT come from then? If it's just everything after the first dash (2.0-20170626.180046-18) with SNAPSHOT, then it might catch something completely random.

damencho:
So here it is, it is green now, just changed the replacement to 2.0.0-SNAPSHOT.
It is currently doing:
`xmlstarlet ed --inplace -u "/_:project/_:dependencies/_:dependency[_:artifactId='ice4j']/_:version" -v 2.0-SNAPSHOT pom.xml`
I will see to get the version first and then replace everything after the first dash before changing it in the pom file.

ibauersachs:
Just replace whatever version is in the target pom (e.g. Videobridge) with what you find in the source pom (i.e. the one from the pull request). There shouldn't be any assumption about the version formats.

damencho:
@ibauersachs just changed it as you suggested, thanks. I will restart these PR tests to try it.
Jenkins, test this please.

damencho:
Works now and with ice4j unit tests :)

PR #130, opened 2017-08-19
remove a transaction if a packet fails validation

bbaldino:
when edge sends ice messages, they send duplicates formatted in the 3489 and 5389 styles until they get a response...whichever one is responded to is the format they'll continue.  sometimes the 3489 packet will be sent first and sometimes the 5389 packet will be sent first.

for some reason, the 3489 packets never pass message integrity checks (i am unable to arrive at the same message integrity value manually when following https://tools.ietf.org/html/rfc3489#section-11.2.8 for a packet edge reproduces, ice4j arrives at the same result i do but edge has something different), so those always result in an error response.  since the duplicate 3489/5389 packets use the same transaction id, if we process the 3489 message first and fail, when we get the 5389 message we'll look up the transaction and just return the previous result (failure), meaning ice will never connect.  if we get the 5389 one first, edge will connect fine.

spoke with @emcho and he suggested just removing transactions upon failure so we won't remember the failed state, which allows us to succeed if the 5389 packet comes second.

### DIFF
--- a/src/main/java/org/ice4j/stack/StunStack.java
+++ b/src/main/java/org/ice4j/stack/StunStack.java
@@ -978,7 +978,8 @@ public class StunStack
             catch(Exception exc)
             {
                 //validation failed. log get lost.
-                logger.log(Level.FINE, "Failed to validate msg: " + ev, exc);
+                logger.log(Level.FINE, "Failed to validate msg, removing transaction: " + ev, exc);
+                removeServerTransaction(sTran);
                 return;
             }
 


bbaldino:
dammit, forgot i had some debug prints, will remove those

ibc:
Excellent catch!  It explains why the issue does not happen in ICE servers that do not manage transactions (but reply stateless to any received STUN binding request).

bgrozev:
Nice catch indeed! The patch looks good to me, but I have one question: does Edge handle a success response *and* a failure response in a way that makes things work?

ibc:
Uppps, good question. If that happens, and if Edge adds custom STUN attributes (such as USER_AGENT, and I think it does), ice4j may just drop those invalid requests. I hope such a hack is not needed...

bbaldino:
@bgrozev i tested that scenario and it does work, so i'm assuming that this is handled by edge (which seems reasonable, given this scheme of 'send both and do the one that is acked successfully')

PR #132, opened 2017-10-19
Add tcptype to Candidate toString method

jalaziz:


jalaziz:
I also noticed `Candidate` doesn't take `tcpType` into account for equals. I didn't want to assume it was safe to add `tcpType` to `equals()`, since it could result in breaking code. However, I believe it's the right thing to do.

jalaziz:
As it turns out, Chrome does not accept the `tcptype` attribute in the SDP string for SSLTCP. This results in some conflicts with Jitsi / ICE4J.

PR #133, opened 2017-11-10
Annotate Google SSLTCP hardcoded values.

guusdk:
I had a difficult time trying to figure out how Google Turn SSL (SSLTCP) candidate works.
My analysis might help future-me (or future-people-like-me), hence this PR.

This commit adds documentation only. It does not introduce functional changes.

The documentation was generated by crafting hex dumps from the values hardcoded
in the source code, which were fed to Wireshark (using its "import from hex dump"
option).

PR #134, opened 2017-11-14
Allow bind to wildcard (any) local address.

guusdk:
This commit allows the project to bind to any local address, by setting a property.

### DIFF
--- a/src/main/java/org/ice4j/ice/harvest/AbstractTcpListener.java
+++ b/src/main/java/org/ice4j/ice/harvest/AbstractTcpListener.java
@@ -406,15 +406,44 @@ public abstract class AbstractTcpListener
     protected void init()
         throws IOException
     {
+        boolean bindWildcard = !StackProperties.getBoolean(
+                StackProperties.BIND_WILDCARD,
+                false);
+
+        // Use a set to filter out any duplicates.
+        Set<InetSocketAddress> addressesToBind = new HashSet<>();
+
         for (TransportAddress transportAddress : localAddresses)
         {
-            ServerSocketChannel channel
-                = MuxServerSocketChannelFactory
-                    .openAndBindMuxServerSocketChannel(
+            addressesToBind.add( new InetSocketAddress(
+                bindWildcard ? null : transportAddress.getAddress(),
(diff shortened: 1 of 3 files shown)


guusdk:
Do you guys have a good way to test this? I've tried, but the snapshot-based versioning that's applied in Jitsi projects made me fail a couple of times (while trying to update this in three-levels-down dependency of my project).

guusdk:
At the time of making this PR, another PR was already merged/accepted: https://github.com/jitsi/ice4j/pull/133 Unsure why jitsi-jenkins again asks for a CLA?

damencho:
The CLA message is comming from a dumb plugin, we just need to whitelist you in every project, cause every project has its own settings.

damencho:
Jenkins, add to whitelist.

src/main/java/org/ice4j/ice/harvest/AbstractTcpListener.java
```diff
@@ -406,15 +406,44 @@ public void close()
     protected void init()
         throws IOException
     {
+        boolean bindWildcard = !StackProperties.getBoolean(
```
bgrozev: Why the negation?

guusdk: That's a mistake. Will fix asap.

src/main/java/org/ice4j/ice/harvest/AbstractUdpListener.java
```diff
@@ -205,8 +205,24 @@ static String getUfrag(byte[] buf, int off, int len)
...
+                                        localAddress.getPort(),
+                                        localAddress.getTransport()
+                                );
+        }
+        else
+        {
+            this.localAddress = localAddress;
+        }
+
+        socket = new DatagramSocket( localAddress );
```
bgrozev: Do you want localAddress or this.localAddress here?

bgrozev:
LGTM (pending the question about localAddress)

guusdk:
Good catch. Now fixed.

bgrozev:
Thanks!

PR #138, opened 2018-03-27
Log msg type

bgrozev:


bgrozev:
Jenkins, test this please

src/main/java/org/ice4j/attribute/MessageIntegrityAttribute.java
```diff
@@ -283,19 +291,24 @@ public void decodeAttributeBody( byte[] attributeValue,
...
         {
             /* attribute part of a response, use the local key */
             key = stunStack.getCredentialsManager().getLocalKey(username);
         }
+        else // msgType is Indication
+        {
+            logger.warning("Can not encode a message of type Indication.");
+        }
 
+        Objects.requireNonNull(key, "key=null; msgType=" + msgType);
```
bbaldino: since we know this will only happen in the else case, is it worth just augmenting the error message logged there instead of having this here?

src/main/java/org/ice4j/message/Message.java
```diff
@@ -1058,7 +1063,7 @@ public static Message decode(byte binMessage[], char offset, char arrayLen)
         int originalOffset = offset;
         arrayLen = (char)Math.min(binMessage.length, arrayLen);
 
-        if(binMessage == null || arrayLen - offset < Message.HEADER_LENGTH)
+        if(arrayLen - offset < Message.HEADER_LENGTH)
```
bbaldino: why do we no longer need the null check?

bgrozev: Because we reference binMessage.length on the line above

src/main/java/org/ice4j/attribute/MessageIntegrityAttribute.java
bgrozev: I don't know whether this only happens in the "else" case, we could get `null` from the credentials manager. Actually, the reason I'm adding these logs is to try and understand why a see an exception in the logs every now and again.

src/main/java/org/ice4j/message/Message.java
bbaldino: 😭

bgrozev:
The test failure is unrelated

Issue #139, opened 2018-05-16
SinglePortUdpHarvester should close / unbind when appropriate.


guusdk:
I'm running ice4j (as part of the Jitsi Videobridge) in a plugin-based environment. This means that users can start and stop ice4j, without (re)starting the JVM.

Upon usage of our component _after it was restarted_, we end up with these messages in our log file:

> org.ice4j.ice.harvest.SinglePortUdpHarvester - Failed to create SinglePortUdpHarvester for address 198.51.100.1:10000/udp: java.net.BindException: Address already in use (Bind failed)

It appears that this harvester does not properly release all resources, when it's stopped.

PR #140, opened 2018-05-16
AbstractUdpListener should be able to release its resources

guusdk:
A new 'close' method has been added to AbstractUdpListener that is modeled after the 'close' implementation of AbstractTcpListener.

This change can be used to prevent the issue described in https://github.com/jitsi/ice4j/issues/139

### DIFF
--- a/src/main/java/org/ice4j/ice/harvest/AbstractUdpListener.java
+++ b/src/main/java/org/ice4j/ice/harvest/AbstractUdpListener.java
@@ -197,6 +197,11 @@ public abstract class AbstractUdpListener
     private final Thread thread;
 
     /**
+     * Triggers the termination of the threads of this instance.
+     */
+    private boolean close = false;
+
+    /**
      * Initializes a new <tt>SinglePortUdpHarvester</tt> instance which is to
      * bind on the specified local address.
      * @param localAddress the address to bind to.
@@ -255,6 +260,15 @@ public abstract class AbstractUdpListener
     }
 
     /**
+     * Triggers the termination of the threads of this instance.
+     */
+    public void close()
+    {
+        close = true;
+        socket.close(); // causes socket#receive to stop blocking.
+    }
+
+    /**
      * Perpetually reads datagrams from {@link #socket} and handles them
      * accordingly.
      *
@@ -271,7 +285,10 @@ public abstract class AbstractUdpListener
 
         do
         {
-            // TODO: implement stopping the thread with a switch?
+            if (close)
+            {
+                break;
+            }
 
             buf = getFreeBuffer();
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/ice4j/ice/harvest/AbstractUdpListener.java
```diff
@@ -316,7 +336,12 @@ private void runInHarvesterThread()
         }
         while (true);
 
-        // TODO we are all done, clean up.
+        //now clean up and exit
+        for (MySocket candidateSocket : sockets.values())
+        {
+            candidateSocket.close();
+        }
+        socket.close();
```
bgrozev: AbstractUdpListener.socket.close() is called here, and also in AbstractUdpListener.close(). Is there a reason?

src/test/java/org/ice4j/ice/harvest/SinglePortUdpHarvesterTest.java
```diff
@@ -0,0 +1,96 @@
+package org.ice4j.ice.harvest;
+
+import org.ice4j.Transport;
+import org.ice4j.TransportAddress;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.net.BindException;
```
bgrozev: Thank you for adding tests! 

Nits: license header, * imports

src/main/java/org/ice4j/ice/harvest/AbstractUdpListener.java
guusdk: I've initially left it out here, but considered that the loop could also be broken due to an unexpected exception. That's why I put the call to `socket.close()` back in. If the socket is already closed, this second invocation is a no-op.

src/main/java/org/ice4j/ice/harvest/AbstractUdpListener.java
```diff
@@ -316,7 +336,12 @@ private void runInHarvesterThread()
         }
         while (true);
 
-        // TODO we are all done, clean up.
+        //now clean up and exit
+        for (MySocket candidateSocket : sockets.values())
+        {
+            candidateSocket.close();
```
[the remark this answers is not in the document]
guusdk: I've not tested this, other than the unit test that I supplied. You're right, this might be causing ConcurrentModificationExceptions. Do you see a more elegant solution than creating a new collection?

src/test/java/org/ice4j/ice/harvest/SinglePortUdpHarvesterTest.java
guusdk: Le sigh (pardon my French).

PR #147, opened 2018-10-03
feat: Makes the use of a component socket optional.

bgrozev:


src/main/java/org/ice4j/ice/CandidatePair.java
```diff
@@ -745,11 +743,42 @@ public Socket getSocket()
...
+        {
+            SocketAddress remoteAddress
+                = remoteCandidate.getTransportAddress();
+            if (remoteAddress != null)
+            {
+                return localCandidate.getCandidateIceSocketWrapper(remoteAddress);
+            }
+        }
+
+        return localCandidate.getCandidateIceSocketWrapper();
```
bbaldino: how does it make sense to have a wrapper but with no remote candidate attached?  will it be updated automatically if/once a remote candidate is determined?

bbaldino:
Went through the bridge and the changes there don't seem terrible, actually, and I think they'd be even cleaner once the rewrite hits.  I'm still curious about the answers to the questions above, but this doesn't seem as bad after looking at jvb.

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -486,12 +488,14 @@ public Component createComponent(  IceMediaStream stream,
...
-                                       int            maxPort,
-                                       KeepAliveStrategy keepAliveStrategy)
+    public Component createComponent(
+        IceMediaStream stream,
+        Transport transport,
+        int preferredPort,
+        int minPort,
+        int maxPort,
+        KeepAliveStrategy keepAliveStrategy,
+        boolean useComponentSocket)
```
ibauersachs: You might want to keep the existing constructor (with a default that resembles the previous behavior). Otherwise this needs a new API version.

src/main/java/org/ice4j/ice/Component.java
```diff
@@ -169,10 +171,12 @@
...
      * @param mediaStream the {@link IceMediaStream} instance that would be the
      * parent of this component.
+     * @param useComponentSocket whether the component socket should be used.
      */
-    protected Component(int            componentID,
+    protected Component(int componentID,
                         IceMediaStream mediaStream,
-                        KeepAliveStrategy keepAliveStrategy)
+                        KeepAliveStrategy keepAliveStrategy,
+                        boolean useComponentSocket)
```
ibauersachs: Same here with keeping the existing constructor and the version.

src/main/java/org/ice4j/ice/CandidatePair.java
bgrozev: I'm not quite sure (this is a rollback from a previous version). But I think if you are using the dynamic ports that you just bind to a local port and it's ready for reading regardless of the remote candidate (though you will need the target if you want to send packets).

bgrozev:
1: Without the "component socket" it falls back to the previous ice4j API and behavior. That is, one candidate pair is selected and nominated, and once this is done (we move to the COMPLETED state) you can not change to another pair. You can access the selected pair through `Component.getSelectedPair()`, and that's where you get your socket.

2 & 3: This is not supported. ICE needs to run again.

I'll do the API changes that Ingo suggested which should also fix the tests.

src/main/java/org/ice4j/ice/Component.java
bgrozev: This one is `protected`. Theoretically, people might extend `Component`, but I highly doubt anyone is doing it. Are there any automated tools that check the API compatibility? Do you still think it's necessary to re-add the constructor with the previous signature?

bgrozev: I added it, it doesn't hurt

ibauersachs: Part of the Maven build should actually check the API compatibility.

PR #150, opened 2018-10-12
Fix thread leak when Agent is freed before termination timer triggered.

mstyura:
When `org.ice4j.TERMINATION_DELAY` is set to some big value to effectively "disable" termination a termination thread is outlive owning Agent, because there was missing thread interruption in `free()`

### DIFF
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -23,6 +23,7 @@ import java.math.*;
 import java.net.*;
 import java.security.*;
 import java.util.*;
+import java.util.concurrent.*;
 import java.util.logging.*;
 
 import org.ice4j.*;
@@ -113,6 +114,36 @@ public class Agent
      */
     public static final String PROPERTY_ICE_PROCESSING_STATE
                                             = "IceProcessingState";
+
+    /**
+     *  The ScheduledExecutorService to execute planned agent termination
+     */
+    private static final ScheduledExecutorService agentTerminationScheduler;
+
+    static {
+        final ScheduledThreadPoolExecutor terminationExecutor =
+            new ScheduledThreadPoolExecutor(0);
+        terminationExecutor.setKeepAliveTime(10, TimeUnit.SECONDS);
+        terminationExecutor.setRemoveOnCancelPolicy(true);
+        agentTerminationScheduler =
+            Executors.unconfigurableScheduledExecutorService(
+                terminationExecutor);
+    }
+
+    /**
+     * Termination task which will be scheduled with timeout
+     */
+    private final Runnable terminationRunnable = new Runnable()
+    {
+        @Override
+        public void run()
+        {
+            terminate(IceProcessingState.TERMINATED);
+            synchronized (terminationFutureSyncRoot) {
+                terminationFuture = null;
+            }
+        }
+    };
 
     /**
      * The LinkedHashMap used to store the media streams
@@ -246,14 +277,14 @@ public class Agent
     private StunStack stunStack;
 
     /**
-     * The thread that we use for moving from COMPLETED into a TERMINATED state.
+     * The scheduled task to terminate this <tt>Agent</tt>
      */
-    private TerminationThread terminationThread;
+    private ScheduledFuture<?> terminationFuture;
 
     /**
-     * The object used to synchronize access to {@link #terminationThread}.
+     * The object used to synchronize access to {@link #terminationFuture}.
      */
-    private final Object terminationThreadSyncRoot = new Object();
+    private final Object terminationFutureSyncRoot = new Object();
 
     /**
      * The thread that we use for STUN keep-alive.
@@ -2150,17 +2181,48 @@ public class Agent
     }
 
     /**
-     * Initializes and starts the {@link TerminationThread}
+     * Initializes and starts the {@link #terminationFuture}
      */
     private void scheduleTermination()
     {
-        synchronized (terminationThreadSyncRoot)
+        /*
+         * RFC 5245 says: Once ICE processing has reached the Completed state for
+         * all peers for media streams using those candidates, the agent SHOULD
+         * wait an additional three seconds, and then it MAY cease responding to
+         * checks or generating triggered checks on that candidate.  It MAY free
+         * the candidate at that time.
+         * <p>
+         * This method is scheduling such a termination.
+         */
+
+        boolean runTerminationImmediately = false;
+
+        synchronized (terminationFutureSyncRoot)
         {
-            if (terminationThread == null)
+            if (terminationFuture == null)
             {
-                terminationThread = new TerminationThread();
-                terminationThread.start();
+                long terminationDelay
+                    = Integer.getInteger(
+                        StackProperties.TERMINATION_DELAY,
+                        DEFAULT_TERMINATION_DELAY);
+
+                if (terminationDelay > 0)
+                {
+                    terminationFuture
+                        = agentTerminationScheduler.schedule(
+                            terminationRunnable,
+                            terminationDelay,
+                            TimeUnit.MILLISECONDS);
+                }
+                else
+                {
+                    runTerminationImmediately = true;
+                }
             }
+        }
+
+        if (runTerminationImmediately) {
+            terminationRunnable.run();
         }
     }
 
@@ -2300,12 +2362,12 @@ public class Agent
 
         // cancel termination timer in case agent is freed
         // before termination timer is triggered
-        synchronized (terminationThreadSyncRoot)
+        synchronized (terminationFutureSyncRoot)
         {
-            if (terminationThread != null)
+            if (terminationFuture != null)
             {
-                terminationThread.interrupt();
-                terminationThread = null;
+                terminationFuture.cancel(true);
+                terminationFuture = null;
             }
         }
 
@@ -2719,63 +2781,5 @@ public class Agent
     protected Logger getLogger()
     {
         return logger;
-    }
-
-    /**
-     * RFC 5245 says: Once ICE processing has reached the Completed state for
-     * all peers for media streams using those candidates, the agent SHOULD
-     * wait an additional three seconds, and then it MAY cease responding to
-     * checks or generating triggered checks on that candidate.  It MAY free
-     * the candidate at that time.
-     * <p>
-     * This <tt>TerminationThread</tt> is scheduling such a termination and
-     * garbage collection in three seconds.
-     */
-    private class TerminationThread
-        extends Thread
-    {
-
-        /**
-         * Creates a new termination timer.
-         */
-        private TerminationThread()
-        {
-            super("TerminationThread");
-        }
-
-        /**
-         * Waits for a period of three seconds (or whatever termination
-         * interval the user has specified) and then moves this <tt>Agent</tt>
-         * into the terminated state and frees all non-nominated candidates.
-         */
-        @Override
-        public synchronized void run()
-        {
-            long terminationDelay
-                = Integer.getInteger(
-                StackProperties.TERMINATION_DELAY,
-                DEFAULT_TERMINATION_DELAY);
-
-            if (terminationDelay >= 0)
-            {
-                try
-                {
-                    wait(terminationDelay);
-                }
-                catch (InterruptedException ie)
-                {
-                    logger.log(Level.FINEST, "Interrupted while waiting. Will "
-                                   + "speed up termination",
-                               ie);
-                }
-            }
-
-            terminate(IceProcessingState.TERMINATED);
-
-            synchronized (terminationThreadSyncRoot)
-            {
-                terminationThread = null;
-            }
-        }
     }
 }
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -120,13 +120,14 @@ public class Agent
      */
     private static final ScheduledExecutorService agentTerminationScheduler;
 
-    static {
-        final ScheduledThreadPoolExecutor terminationExecutor =
-            new ScheduledThreadPoolExecutor(0);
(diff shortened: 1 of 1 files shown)


mstyura:
@jitsi-jenkins I've already signed CLA for Individuals with email used in git's commit author field.

mstyura:
@bgrozev could you please review this pull request?

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -2298,6 +2298,17 @@ public void free()
         if (stunKeepAliveThread != null)
             stunKeepAliveThread.interrupt();
 
+        // cancel termination timer in case agent is freed
+        // before termination timer is triggered
+        synchronized (terminationThreadSyncRoot)
```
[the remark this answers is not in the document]
mstyura: >> If the thread is still alive, it is most probably in the call to wait. If that's the case, interrupting it will result in it executing terminate(TERMINATED), which is probably not what we want.

Yes, but I didn't consider it as serious issue, since Agent is anyway freeing. Actually I have a better version of this fix which does not have this problem, but I wanted to start with smaller pull request.

>> What is the problem you're trying to solve? This doesn't solve the problem of terminate executing after the agent is freed, and the leak should only last 3 seconds at most (the default termination delay).

The problem I'm trying to solve is described in my [question](https://community.jitsi.org/t/webrtc-client-to-jvb-connection-recovery/15418) for Scenario 2.
I basically set delay timeout to "infinite" value via configuration to effectively cancel termination, it caused thread to live "forever" in waiting state and it probably also keep reference to dead Agent.

I'll update this pull request with better version of fix, which does proper cancellation as well as saves a lot of threads for my scenario by using scheduled thread pool execution service to execute delayed termination.

mstyura:
Updated pull request with better version of fix which:
1. Does not create thread per agent to schedule termination
2. Does support "infinite" delay without leaking treads
3. Does proper cancellation of termination when agent is freed before termination due to "infinite" delay.

bgrozev:
Just for context, can you please explain what the problem you're trying to solve is?

paweldomas:
Hey @bgrozev, you can find more context here:
https://community.jitsi.org/t/webrtc-client-to-jvb-connection-recovery/15418/4

bgrozev:
jenkins add to whitelist

mstyura:
@bgrozev is there anything I can do to make this pull request accepted soon?

src/main/java/org/ice4j/ice/Agent.java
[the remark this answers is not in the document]
paweldomas: I'd say that for the infinite termination feature this thread should not be started at all.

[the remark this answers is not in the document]
mstyura: By "infinite" I meant something like day or week (I believe no peer will connect longer in single session) :) But we could also define some value as infinite and do not start delayed termination task at all. BTW, there is more recent version of this pull request which does not create threads explicitly at all.

paweldomas:
@mstyura will there be another PR for the infinite termination or do you plan on solving that just by setting very large delay (day or a week) ? I'd rather have an explicit "infinite" delay lets say -1, then 0 - would mean run immediately and finally the other values are the actual delay time. @bgrozev WDYT ?

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -114,6 +115,36 @@
...
                                             = "IceProcessingState";
 
+    /**
+     *  The ScheduledExecutorService to execute planned agent termination
+     */
+    private static final ScheduledExecutorService agentTerminationScheduler;
+
+    static {
+        final ScheduledThreadPoolExecutor terminationExecutor =
+            new ScheduledThreadPoolExecutor(0);
```
paweldomas: I'm reading [here] that 0 is not a good idea:
>Additionally, it is almost never a good idea to set corePoolSize to zero or use allowCoreThreadTimeOut because this may leave the pool without threads to handle tasks once they become eligible to run.
I think it would be reasonable to allow more than one termination task to execute at the same time. @bgrozev WDYT ?
[here]: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html#ScheduledThreadPoolExecutor(int)

mstyura: @paweldomas I've also read that. But it is occurred to be inconsistency in doc... I'll find prooflink and attach. I've checked both in test application and inside JVB that it works as expected - create threads on demand and release them after timeout of no work.

mstyura: @paweldomas here is a link https://bugs.java.com/bugdatabase/view_bug.do?bug_id=7091003

mstyura:
@paweldomas I could also implement explicit "infinite" delay value, but regarding negative values - old implementation was terminate immediately when `delay < 0`, but if we change it to -1 some client's of library might not expect that Agent is not terminated at all with updated implementation.
If possible I prefer merging this pull request without adding explicit infinite delay, but then I'd like to create new pull request with explicit infinite delay implemented the way we decided.

paweldomas:
Yes, maybe lets not break old clients. Negative should still mean immediately like you wrote then.

src/main/java/org/ice4j/ice/Agent.java
mstyura: >I think it would be reasonable to allow more than one termination task to execute at the same time. 

@paweldomas this scheduled thread pool is not limited to one thread (it actually has no limit on threads to create), so many termination tasks from different agents could be executed in parallel. But using single shared thread pool between all agents results in using as low number of threads as possible.

mstyura: > So with pool size 0 there will be no threads kept alive, but does it allow to execute multiple tasks at the same time ? Do I understand it correctly ?

@paweldomas this `corePoolSize` which is set to `0` means the number of threads in pool to keep even when there is no work to execute. So, since it is a library, I decided that their users don't want to pay some threads when library is not actively using. I configured executor to not too keep "warm" (core) threads if there is no more for more than 10 seconds.


So basically, when pool is created it's empty.
When some task is queued - thread is allocated (if there is not any).
Work is executed.
Thread has no more work for 10 seconds.
Thread is released from pool.
Pool is empty again, because no work to execute.

bgrozev:
I'm sorry, but I haven't had the chance to look at this. Does delaying the termination allow for new candidate pairs to be discovered/succeed? I was expecting that this requires much more work, and if all we're doing is keeping the existing pairs alive we should use KeepAliveStrategy.ALL_SUCCEEDED.

I'm probably missing something, and I'll look at the PR more carefully in the coming weeks (very busy right now).

mstyura:
> I'm probably missing something, and I'll look at the PR more carefully in the coming weeks (very busy right now).

@bgrozev can we consider this PR as a performance/memory optimization due to reducing amount of threads created during ICE Agent lifetime? In such case it does not look such scary and actually does not change externally observable behavior at all - it only reduces amount of idle threads. 
And hence it is so simple and does not change behavior - it could be merged sooner, without so much careful review, and all of the other not related to performance benefits can be investigated later? 
I actually blocked by absence of these in ice4j, because now I running out of threads handles on server after some time.

> Does delaying the termination allow for new candidate pairs to be discovered/succeed?

Yes, that what I observe between my client app and JVB with fixed ice4j. 
To check this yourself, you need a client app with continual ICE gathering enabled in WebRTC. In that case WebRTC sends STUN ping from all candidates it discovers during lifetime of peer connection, even they discovered 5 minutes after call started. During such call some candidates removed, some added when user switches between networks.

What happen currently when termination of agent is done after 3 second? 
In my app I don't report any client candidates to JVB from client.
JVB is has single candidate with publicly available static IP address.
Maybe not reporting candidates from client is already something bad done in my app, because if candidates is not discovered during 3 seconds after first successful candidate pair discovered - it will not be discovered at all currently.
So, in my case JVB discovers client candidates by receiving STUN pings from client. 
When agent is not terminated and STUN ping is received - new client candidate is discovered or existing candidate confirmed "alive".
When candidate is discovered before termination of ice Agent, JVB sends STUN ping back to client and receive STUN ping response.
After this STUN handshake - client candidate considered "authorized" and added into [`ComponentSocket.authorizedAddresses`](https://github.com/jitsi/ice4j/blob/master/src/main/java/org/ice4j/ice/ComponentSocket.java#L54).
When data packet is received from authorized socket it passed further to JVB which will broadcast it to other peers.

So, when agent is terminated and client discovers new candidate and send STUN ping to JVB - the ice Agent does respond with STUN ping response, but it does not send STUN ping back to client. In that case STUN ping "handshake" is never completed for candidate pair and hence newly discovered client address is not added to [`ComponentSocket.authorizedAddresses`](https://github.com/jitsi/ice4j/blob/master/src/main/java/org/ice4j/ice/ComponentSocket.java#L54).
When WebRTC starts sending data from new address, because it thinks it is writable (because of STUN ping response from JVB!), the data which is arrived to socked at JVB side is discarded, because it came from non-authorized address. So, when agent is terminated no actual data is transmitted between JVB and client on newly discovered candidate.


>if all we're doing is keeping the existing pairs alive we should use KeepAliveStrategy.ALL_SUCCEEDED.

No that is not enough, because it will only help (but I think it is actually no op, because terminated agent responds to STUN ping) for cases when user joined conference with multiple networks available and hence multiple candidates, and this candidates/interfaces does not changed during call, which is not valid in case of mobile app switching between WiFi/3G networks.

paweldomas:
I think we can treat this PR just as an optimisation.

Anyway there are few styling issues which I will try to comment in the file inline.

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -114,6 +115,36 @@
     public static final String PROPERTY_ICE_PROCESSING_STATE
                                             = "IceProcessingState";
 
+    /**
+     *  The ScheduledExecutorService to execute planned agent termination
+     */
+    private static final ScheduledExecutorService agentTerminationScheduler;
+
+    static {
```
paweldomas: `{` should go on the next line (in all other places as well)

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -114,6 +115,36 @@
...
+        final ScheduledThreadPoolExecutor terminationExecutor =
```
paweldomas: the `=` sign should be on the next line

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -114,6 +115,36 @@
...
+    /**
+     * Termination task which will be scheduled with timeout
+     */
+    private final Runnable terminationRunnable = new Runnable()
+    {
+        @Override
+        public void run()
+        {
+            terminate(IceProcessingState.TERMINATED);
+            synchronized (terminationFutureSyncRoot) {
```
paweldomas: the `{` should be on the next line

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -2150,18 +2181,49 @@ protected long calculateStunConnCheckRTO()
...
+        /*
+         * RFC 5245 says: Once ICE processing has reached the Completed state for
+         * all peers for media streams using those candidates, the agent SHOULD
+         * wait an additional three seconds, and then it MAY cease responding to
+         * checks or generating triggered checks on that candidate.  It MAY free
+         * the candidate at that time.
+         * <p>
+         * This method is scheduling such a termination.
+         */
+
```
paweldomas: unnecessary extra line

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -2150,18 +2181,49 @@ protected long calculateStunConnCheckRTO()
...
+                            TimeUnit.MILLISECONDS);
+                }
+                else
+                {
+                    runTerminationImmediately = true;
+                }
             }
         }
+
+        if (runTerminationImmediately) {
```
paweldomas: `{` should be on the next line

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -114,6 +115,36 @@
...
+     *  The ScheduledExecutorService to execute planned agent termination
+     */
+    private static final ScheduledExecutorService agentTerminationScheduler;
+
+    static {
+        final ScheduledThreadPoolExecutor terminationExecutor =
+            new ScheduledThreadPoolExecutor(0);
+        terminationExecutor.setKeepAliveTime(10, TimeUnit.SECONDS);
+        terminationExecutor.setRemoveOnCancelPolicy(true);
+        agentTerminationScheduler =
```
paweldomas: the `=` sign should be on the next line

mstyura:
@paweldomas thank you very much for comments, I've updated formatting. 
Is there existing config for IDE with preferred code style in ice4j and rest of Jitsi projects?

paweldomas:
We used to have one for the Jitsi desktop project, but it doesn't seem to be accessible anymore, sorry.

paweldomas:
Yeah, the tests are fine. This failure is related to a physical iOS device which I need to fix. I'm going to disable the iOS tests for now.

paweldomas:
I'm going to merge the PR after checking with @bgrozev

damencho:
> @paweldomas thank you very much for comments, I've updated formatting.
> Is there existing config for IDE with preferred code style in ice4j and rest of Jitsi projects?

At  the bottom of the page there is an eclipse formatter: https://desktop.jitsi.org/Documentation/CodeConvention

The link was broken as Pawel mentioned, but I just fixed it.

paweldomas:
ok now we have to bubble up this change to the JVB

mstyura:
@paweldomas I'll do a pull request to JVB soon

mstyura:
https://github.com/jitsi/jitsi-videobridge/pull/725 done

PR #151, opened 2018-10-29
Reduce number of threads created by ice4j. Eliminate StunKeepAliveThread.

mstyura:
Replaced manual scheduling of repeated Stun Keep Alive with `ScheduledExecutorService`.
Observed enhancement in my environment: 30 conferences with 3 audio only peers each.
**Before**: `90` instances of `StunKeepAliveThread`
**After**: single `Thread` instance in thread pool which handles both termination & stun keep alive.

### DIFF
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -132,6 +132,10 @@ public class Agent
             {
                 Thread thread = defaultThreadFactory.newThread(r);
                 thread.setName("ice4j.Agent-" + thread.getName());
+                if (!thread.isDaemon())
+                {
+                    thread.setDaemon(true);
+                }
                 return thread;
             }
         };
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -116,17 +116,31 @@ public class Agent
                                             = "IceProcessingState";
 
     /**
-     *  The ScheduledExecutorService to execute planned agent termination
+     *  The ScheduledExecutorService to execute Agent's scheduled tasks
      */
-    private static final ScheduledExecutorService agentTerminationScheduler;
+    private static final ScheduledExecutorService agentTasksScheduler;
 
     static
     {
+        final ThreadFactory agentThreadFactory = new ThreadFactory()
+        {
+            private final ThreadFactory defaultThreadFactory
+                = Executors.defaultThreadFactory();
+
+            @Override
+            public Thread newThread(Runnable r)
+            {
+                Thread thread = defaultThreadFactory.newThread(r);
+                thread.setName("ice4j.Agent-" + thread.getName());
+                return thread;
+            }
+        };
+
         final ScheduledThreadPoolExecutor terminationExecutor
-            = new ScheduledThreadPoolExecutor(0);
+            = new ScheduledThreadPoolExecutor(0, agentThreadFactory);
         terminationExecutor.setKeepAliveTime(10, TimeUnit.SECONDS);
         terminationExecutor.setRemoveOnCancelPolicy(true);
-        agentTerminationScheduler
+        agentTasksScheduler
             = Executors.unconfigurableScheduledExecutorService(
                 terminationExecutor);
     }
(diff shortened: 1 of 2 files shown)


mstyura:
@paweldomas could you please review?

mstyura:
Contrasts with [acquisition](https://community.jitsi.org/t/jitsi-and-8x8/15538), but hope it will not prevent this and further my pull requests merged soon.

paweldomas:
Hey @mstyura I'm on vacation right now and will be coming back on the 1st, but will try to take a look sooner.

mstyura:
Have a good vocation @paweldomas! If you have someone on your mind who could review this PR as well - that's a good option to save your vocation :)

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -2441,80 +2547,14 @@ public void setGeneration(int generation)
...
     /**
-     * Determines whether {@link #runInStunKeepAliveThread()} is to run.
+     * Determines whether {@link #stunKeepAliveRunnable} is to run.
      *
-     * @return <tt>true</tt> if <tt>runInStunKeepAliveThread()</tt> is to run;
+     * @return <tt>true</tt> if <tt>{@link #stunKeepAliveRunnable}</tt> is to run;
      * otherwise, <tt>false</tt>
      */
-    private boolean runInStunKeepAliveThreadCondition()
+    private boolean runStunKeepAliveCondition()
```
bbaldino: if we're gonna rename this, i'd prefer something more like `shouldRunStunKeepAlive`

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -116,17 +116,35 @@
...
     /**
-     *  The ScheduledExecutorService to execute planned agent termination
+     *  The ScheduledExecutorService to execute Agent's scheduled tasks
      */
-    private static final ScheduledExecutorService agentTerminationScheduler;
+    private static final ScheduledExecutorService agentTasksScheduler;
 
     static
     {
+        final ThreadFactory agentThreadFactory = new ThreadFactory()
```
bbaldino: a more generic `ThreadFactory` class which takes in a name prefix and whether or not threads should be daemon threads might be useful.

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -147,6 +165,75 @@ public void run()
...
+        {
+            if (!runStunKeepAliveCondition())
+            {
+                synchronized (stunKeepAliveFutureSyncRoot)
+                {
+                    if (!runStunKeepAliveCondition())
+                    {
+                        if (stunKeepAliveFuture != null)
+                        {
+                            stunKeepAliveFuture.cancel(false);
```
bbaldino: we've got this code in two places, maybe worth moving to a helper?

mstyura:
@bbaldino thank you very much for your review. I've tried to fix what you've asked. The major change since reviewed state is that instead of anonymous runnable I've extracted dedicated class which support schedule/cancel/shouldSchedule - so everything related to STUN keep-alives now coupled in same class.

src/main/java/org/ice4j/util/CustomizableThreadFactory.java
```diff
@@ -0,0 +1,52 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ice4j.util;
+
+import java.util.concurrent.*;
+
+public final class CustomizableThreadFactory implements ThreadFactory
```
bbaldino: can you throw a quick javadoc on here?

mstyura:
Crossing fingers and hope @bgrozev will have a look at this soon, because I already have several pending PR on top of this :)

bgrozev:
Thanks for checking with me. I think that's a nice optimization, and the code LGTM.


Just informational: we have this utility in libjitsi that does something very similar to CustomizableThreadFactory:
https://github.com/jitsi/libjitsi/blob/master/src/org/jitsi/util/ExecutorUtils.java#L45

Once we figure out what we want to do with libjitsi's dependency on ice4j we might want to remove the duplication (I'd suggest reusing CustomizableThreadFactory in lj).

PR #152, opened 2018-10-30
Reduce number of threads created by ice4j. Eliminate ICE PaceMaker thread & Timer.

mstyura:
Replaced manual recurrent `Thread`-based scheduling of `CheckList` checks with `ScheduledExecutorService `.
Replaced `Timer`, which under the hood creates `Thread` with `ScheduledExecutorService `.

Observed enhancement in my environment: 30 conferences with 3 audio only peers each.
**Before**: **181** instances of `ICE PaceMaker` thread.
**After**: **1** `Thread` instance inside thread pool which handles termination & stun keep alives & ICE PaceMaker.

### DIFF
--- a/src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
+++ b/src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
@@ -19,6 +19,8 @@ package org.ice4j.ice;
 
 import java.net.*;
 import java.util.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.*;
 import java.util.logging.*;
 
 import org.ice4j.*;
@@ -55,6 +57,11 @@ class ConnectivityCheckClient
     private final Agent parentAgent;
 
     /**
+     * A scheduled executor service to perform scheduled task of the client
+     */
+    private final ScheduledExecutorService scheduledExecutorService;
+
+    /**
      * The <tt>StunStack</tt> that we will use for connectivity checks.
      */
     private final StunStack stunStack;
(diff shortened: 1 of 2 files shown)


mstyura:
@bbaldino could you please review these changes?

src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
```diff
@@ -907,118 +1009,19 @@ private long getNextWaitInterval()
...
     }
 
     /**
      * Stops and removes all <tt>PaceMaker</tt>s.
      */
     public void stop()
     {
-        synchronized (paceMakers)
+        Iterator<PaceMaker> paceMakersIter = paceMakers.iterator();
+        while(paceMakersIter.hasNext())
```
bbaldino: if i'm reading the [docs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html) correctly, i think we could run into an issue here.  the iterator will be valid (no concurrent modification issues), but, the view is only consistent with when the iterator is created, so if a pacemaker was added after we created the iterator here we won't see it and shut it down.

src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
```diff
@@ -847,23 +866,91 @@ public void processTimeout(StunTimeoutEvent ev)
...
+            {
+                if (cancelled.get())
+                {
+                    return;
+                }
+
+                CandidatePair pairToCheck = checkList.popTriggeredCheck();
+
+                //if there are no triggered checks, go for an ordinary one.
+                if (pairToCheck == null)
```
bbaldino: add braces here please

src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
```diff
@@ -847,23 +866,91 @@ public void processTimeout(StunTimeoutEvent ev)
...
+                            = startCheckForPair(pairToCheck);
+
+                        if (transactionID == null)
+                        {
+                            logger.info(
+                                "Pair failed: "
+                                    + pairToCheck.toShortString());
+                            pairToCheck.setStateFailed();
+                        }
+                        else
```
bbaldino: braces

src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
mstyura: Done

mstyura:
@bbaldino thank you very much for the review, I've done style changes as you requested. If no more approves are required I'd like to this one be merged, because I have next on top of this PR to review :)

PR #153, opened 2018-10-31
Stun client transaction re-transmission based on `ScheduledThreadPoolExecutor`

mstyura:
Re-implemented STUN Client transaction re-transmission based on `ScheduledThreadPoolExecutor`.
This allowed to avoid usage of complicated synchronization primitives and enhanced thread's re-use in the thread pool.
It also fixed test `testMaxWaitIntervalConfigurationParameter` in `TransactionSupportTests.java` which counted number of re-transmission attempt. 
Previous implementation had bug when doubling re-transmission delay - it exceed max allowed delay.

Observed enhancement in my environment: `30` conferences with `3` audio only peers each.
**Before**: **233** threads inside `StunClientTransaction` thread pool.
**After**: **1** Thread instance inside `StunClientTransaction` thread pool.

### DIFF
--- a/src/main/java/org/ice4j/stack/StunClientTransaction.java
+++ b/src/main/java/org/ice4j/stack/StunClientTransaction.java
@@ -20,7 +20,6 @@ package org.ice4j.stack;
 import java.io.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
-import java.util.concurrent.locks.*;
 import java.util.logging.*;
 
 import org.ice4j.*;
@@ -291,7 +290,7 @@ public class StunClientTransaction
         // code just checks whether it has become true.
         cancelled.set(true);
 
-        this.retransmitter.cancel(false);
+        this.retransmitter.cancel();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mstyura:
@bbaldino could you please review these changes?

src/main/java/org/ice4j/stack/StunClientTransaction.java
```diff
@@ -486,27 +301,21 @@ void cancel()
...
-                            stackCallback,
-                            evt.getRawMessage(),
-                            (Response) evt.getMessage(),
-                            request,
-                            transactionID));
-        }
-        finally
+        TransactionID transactionID = getTransactionID();
+
+        logger.log(Level.FINE, "handleResponse tid " + transactionID);
```
bbaldino: maybe change this to a direct call since you're tweaking it (`logger.fine` i think?)

src/main/java/org/ice4j/stack/StunClientTransaction.java
```diff
@@ -486,27 +301,21 @@ void cancel()
...
+        if(!Boolean.getBoolean(StackProperties.KEEP_CRANS_AFTER_A_RESPONSE))
         {
-            lock.unlock();
+            cancel();
```
bbaldino: surprised we don't early return here, but i see the original code didn't so i guess best to keep it that way

src/main/java/org/ice4j/stack/StunClientTransaction.java
```diff
@@ -486,27 +301,21 @@ void cancel()
...
-            responseCollector.processResponse(
-                    new StunResponseEvent(
-                            stackCallback,
-                            evt.getRawMessage(),
-                            (Response) evt.getMessage(),
-                            request,
-                            transactionID));
-        }
-        finally
+        TransactionID transactionID = getTransactionID();
```
bbaldino: do we need to cancel a potentially-pending timedout task scheduled by the retransmitter somewhere here?

src/main/java/org/ice4j/stack/StunClientTransaction.java
```diff
@@ -46,13 +47,12 @@
...
  */
 public class StunClientTransaction
-    implements Runnable
 {
     /**
      * Our class logger.
      */
-    private static final Logger logger
-        = Logger.getLogger(StunClientTransaction.class.getName());
+    private static final java.util.logging.Logger logger
```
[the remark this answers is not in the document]
mstyura: @bbaldino yes, that's due to import of `import org.ice4j.util.*;` which is required for `CustomizableThreadFactory`.

src/main/java/org/ice4j/stack/StunClientTransaction.java
mstyura: @bbaldino it looks like this method is kind of "interceptor" of responses, what it does - it stops re-transmission if it is configured to stop on receiving response and it passed response to actual handler who cares about it.

src/main/java/org/ice4j/stack/StunClientTransaction.java
bbaldino: ah oops, yes i missed that.

src/main/java/org/ice4j/stack/StunClientTransaction.java
bbaldino: yup, i see now

mstyura:
@bbaldino could you please merge this if no other approvals needed?

bbaldino:
was just waiting for the tests to pass

PR #156, opened 2018-11-10
Do not limit org.ice4j.ice.harvest.DISABLE_LINK_LOCAL_ADDRESSES to IPv6.

mstyura:
When link local addresses disabled via setting `org.ice4j.ice.harvest.DISABLE_LINK_LOCAL_ADDRESSES = true` it only disabled IPv6 link-local addresses. 
On some environments like Mac and Windows (with virtual Hyper-V networks adapters) there are link local network interfaces with IPv4 addresses, but they can not be filtered via `org.ice4j.ice.harvest.DISABLE_LINK_LOCAL_ADDRESSES `.
This PR fixes this and filter both IPv6 and IPv6 local addresses.
Relevant brief discussion on [community.jitsi.org](https://community.jitsi.org/t/why-ipv4-link-local-addresses-are-not-disabled-by-disable-link-local-addresses-setting/15683/)

### DIFF
--- a/src/main/java/org/ice4j/StackProperties.java
+++ b/src/main/java/org/ice4j/StackProperties.java
@@ -220,7 +220,7 @@ public class StackProperties
             = "org.ice4j.ice.harvest.ALLOWED_ADDRESSES";
 
     /**
-     * The name of the property which, if set to true, specifies that IPv6
+     * The name of the property which, if set to true, specifies that
      * link local addresses should not be used for candidate allocations.
      */
     public static final String DISABLE_LINK_LOCAL_ADDRESSES
--- a/src/main/java/org/ice4j/ice/harvest/AbstractTcpListener.java
+++ b/src/main/java/org/ice4j/ice/harvest/AbstractTcpListener.java
@@ -286,7 +286,7 @@ public abstract class AbstractTcpListener
         boolean useIPv6 = !StackProperties.getBoolean(
                 StackProperties.DISABLE_IPv6,
                 false);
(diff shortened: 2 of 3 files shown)


mstyura:
@saghul could you please review?

saghul:
LGTM, but I think @bbaldino or @bgrozev are more qualified to judge here.

saghul:
FWIW, I’ve seen iOS use link local addresses when connecting in P2P mode with a desktop machine. Now, that said, if those addresses were not included it would have worked all the same with the actual IP of the default interface, as it does often.

So, my 2 cents are that we should merge and revert if there are major issues, but I can’t foresee any at the moment.

mstyura:
@saghul thank you very much for support. I do like your approach with try and revert (in this particular case I believe risk is low), such way iterations will go faster and we could make Jitsi better and better quickly.

mstyura:
> 
> 
> I'm just a bit more cautious because ice4j is a very fundamental library used by a lot of code, so an issue here will affect a lot of components.

this actually is far less important PR for me compared to all others, so we could postpone it until add separate switch for `IPv4` addresses, if making sure nothing broken is so important.

bgrozev:
I would go and update the documentation for the property in StackProperties.java, since it correctly describes the current (weird) behavior.

mstyura:
@bgrozev then I'm removing "WIP" and it can be merged if it is ok

mstyura:
Removed `IPv6` from comment for `DISABLE_LINK_LOCAL_ADDRESSES`.

PR #157, opened 2018-11-11
Do not explicitly create threads in PacketQueue

mstyura:
Do not explicitly create threads in `PacketQueue`, but borrow threads for execution from `ExecutorService`.

Currently there is one inheritor from `ice4j`'s `PacketQuque<T>` which is [`RawPacketQueue`](https://github.com/jitsi/libjitsi/blob/6c0a31d6193dcd2b012ae3ced656ad403403d9ff/src/org/jitsi/util/RawPacketQueue.java#L21) in `libjitsi` used by [`SctpConnection`](https://github.com/jitsi/jitsi-videobridge/blob/56d941ae8634a2f2d781f00a7d1a259ee7d1d9d6/src/main/java/org/jitsi/videobridge/SctpConnection.java#L255) in videobridge.

It is observed that currently there is `N` threads created to process queues created by `SctpConnection`. `N` is the total number of peers connected to `JVB`. Most of the time threads are idle and blocked in `wait(100)` inside `while(true)` loop inside empty `PacketQueue`.

This pull request enhances implementation of `PacketQueue` which now temporary borrows execution thread from shared `ExecutorService` and return thread back to pool when there is no work, this results in less threads used and better usage of existing threads managed by `ExecutorService`.

As comment at the header of `PacketQueue` states:
```/**
 * An abstract queue of packets. This is meant to eventually be able to be used
 * in the following classes (in ice4j and libjitsi) in place of their ad-hoc
 * queue implementations (which is the reason the class is parameterized):
 *     <br> RTPConnectorOutputStream.Queue
 *     <br> PushSourceStreamImpl#readQ
 *     <br> OutputDataStreamImpl#writeQ
 *     <br> SinglePortHarvester.MySocket#queue
 *     <br> MultiplexingSocket#received (and the rest of Multiplex* classes).
 *
 * @author Boris Grozev
 */ 
```

This `PacketQueue` implementation could also dramatically reduce thread usage and enhance overall system performance when aforementioned classes migrated from ad-hoc queue implementations to `PacketQueue`.

Current implementation allow switching between different `ExecutorService` implementations: depending on actual handler work, different `ExecutorService` implementation might give different overall performance.

Summary:

* Implementation temporrary borrow execution thread from configurable `ExecutorService` instead of explicit thread creation and waiting when no items in queue to proceed.
* Throttling functionality was added to satisfy requirements of future move of [`RTPConnectorOutputStream.Queue`](https://github.com/jitsi/libjitsi/blob/6c0a31d6193dcd2b012ae3ced656ad403403d9ff/src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java#L726) to `PacketQueue`
* Due to `ExecutorService` is usually shared between many queues the support of cooperative multi-tasking was added (queue temporary stops it's processing to give execution time to other queues which might work on same `ExecutorService`)
* Unit tests of `PacketQueue` was added
* [Benchmark](https://github.com/jitsi/ice4j/pull/157/files#diff-1519b8e3ad85071f4ff7c1f77cbdde58R14) to compare different `ExecutorService` performance was added. The comparison with baseline implementation added as en example to [different branch in my fork](https://github.com/mstyura/ice4j/commit/37c28b2ae33dbd940d57e581172e3c36f9530585#diff-1519b8e3ad85071f4ff7c1f77cbdde58R134).

Some of micro-benchmark results on my machine (there are too many variations of `numberOfQueues`/`numberOfItemsInQueue`/`singleQueueItemProcessingWeight` to provide all of them):

|change set|PR #157|
|-|-|
|OS |Ubuntu 18.04, 4.13.0-46-generic
|CPU | i7-4790 CPU
|RAM | 16 Gb
|[iterations](https://github.com/jitsi/ice4j/blob/635d8b98d3cb8d5405dce7a434eb8103727a8ccb/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L24) | 100
|[numberOfQueues](https://github.com/jitsi/ice4j/blob/e466aa49f1009ed00981924fbbd33c5156704571/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L30) | **1200**
|[numberOfItemsInQueue](https://github.com/jitsi/ice4j/blob/e466aa49f1009ed00981924fbbd33c5156704571/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L37) | 1000
|[singleQueueItemProcessingWeight](https://github.com/jitsi/ice4j/blob/e466aa49f1009ed00981924fbbd33c5156704571/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L42) | 300
|[ThreadPerQueuePool](https://github.com/jitsi/ice4j/blob/635d8b98d3cb8d5405dce7a434eb8103727a8ccb/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L56)| avg = 1371 ms, std_dev = 1 ms
|[CachedThreadPerQueuePool](https://github.com/jitsi/ice4j/blob/635d8b98d3cb8d5405dce7a434eb8103727a8ccb/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L77)| avg = 1271 ms, std_dev = 0 ms
|[FixedSizeCPUBoundPool](https://github.com/jitsi/ice4j/blob/635d8b98d3cb8d5405dce7a434eb8103727a8ccb/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L98)| avg = 1229 ms, std_dev = 0 ms
|[ForkJoinCPUBoundPool](https://github.com/jitsi/ice4j/blob/635d8b98d3cb8d5405dce7a434eb8103727a8ccb/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L126)| avg = 1223 ms, std_dev = 0 ms
|[PacketQueueOld](https://github.com/mstyura/ice4j/blob/do-no-block-threads-inside-PacketsQueue-with-PacketQueueOld-benchmark/src/test/java/org/ice4j/util/PacketQueueBenchmarkTests.java#L138)| avg = 1427 ms, std_dev = 1 ms

|change set|PR #157|
|-|-|
|OS |Ubuntu 18.04, 4.13.0-46-generic
|CPU | i7-4790 CPU
|RAM | 16 Gb
|iterations | 100
|numberOfQueues | **800**
|numberOfItemsInQueue | 1000
|singleQueueItemProcessingWeight | 300
|ThreadPerQueuePool| avg = 925 ms, std_dev = 2 ms
|CachedThreadPerQueuePool| avg = 846 ms, std_dev = 0 ms
|FixedSizeCPUBoundPool| avg = 821 ms, std_dev = 0 ms
|ForkJoinCPUBoundPool| avg = 763 ms, std_dev = 0 ms
|PacketQueueOld| avg = 923 ms, std_dev = 1 ms | # probably faster than `ThreadPerQueuePool` here because all threads created eagerly, instead of lazily by `ExecutorService`.

|change set|PR #157|
|-|-|
|OS |Ubuntu 18.04, 4.13.0-46-generic
|CPU | i7-4790 CPU
|RAM | 16 Gb
|iterations | 100
|numberOfQueues | **400**
|numberOfItemsInQueue | 1000
|singleQueueItemProcessingWeight | 300
|ThreadPerQueuePool| avg = 446 ms, std_dev = 0 ms
|CachedThreadPerQueuePool| avg = 422 ms, std_dev = 0 ms
|FixedSizeCPUBoundPool| avg = 413 ms, std_dev = 0 ms
|ForkJoinCPUBoundPool| avg = 411 ms, std_dev = 0 ms
|PacketQueueOld| avg = 473 ms, std_dev = 1 ms

|change set|PR #157|
|-|-|
|OS |Ubuntu 18.04, 4.13.0-46-generic
|CPU | i7-4790 CPU
|RAM | 16 Gb
|iterations | 100
|numberOfQueues | **200**
|numberOfItemsInQueue | 1000
|singleQueueItemProcessingWeight | 300
|ThreadPerQueuePool| avg = 238 ms, std_dev = 0 ms
|CachedThreadPerQueuePool| avg = 191 ms, std_dev = 0 ms
|FixedSizeCPUBoundPool| avg = 207 ms, std_dev = 0 ms
|ForkJoinCPUBoundPool| avg = 209 ms, std_dev = 0 ms
|PacketQueueOld| avg = 240 ms, std_dev = 1 ms

|change set|PR #157|
|-|-|
|OS |Ubuntu 18.04, 4.13.0-46-generic
|CPU | i7-4790 CPU
|RAM | 16 Gb
|iterations | 100
|numberOfQueues | **100**
|numberOfItemsInQueue | 1000
|singleQueueItemProcessingWeight | 300
|ThreadPerQueuePool| avg = 121 ms, std_dev = 0 ms
|CachedThreadPerQueuePool| avg = 113 ms, std_dev = 0 ms
|FixedSizeCPUBoundPool| avg = 98 ms, std_dev = 0 ms
|ForkJoinCPUBoundPool| avg = 106 ms, std_dev = 0 ms
|PacketQueueOld| avg = 122 ms, std_dev = 0 ms

|change set|PR #157|
|-|-|
|OS |Ubuntu 18.04, 4.13.0-46-generic
|CPU | i7-4790 CPU
|RAM | 16 Gb
|iterations | 100
|numberOfQueues | **5000**
|numberOfItemsInQueue | **400**
|singleQueueItemProcessingWeight | 300
|ThreadPerQueuePool| avg = 2196 ms, std_dev = 17 ms
|CachedThreadPerQueuePool| avg = 2179 ms, std_dev = 2 ms
|FixedSizeCPUBoundPool| avg = 2057 ms, std_dev = 0 ms
|ForkJoinCPUBoundPool| avg = 1900 ms, std_dev = 0 ms
|PacketQueueOld| avg = 2541 ms, std_dev = 12 ms

### DIFF
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -114,15 +114,15 @@ public final class AsyncQueueHandler<T>
             {
                 T item;
 
+                if (maxSequentiallyHandledItems > 0 &&
+                    sequentiallyHandledItems >= maxSequentiallyHandledItems)
+                {
+                    onYield();
+                    return;
+                }
+
                 synchronized (syncRoot)
                 {
-                    if (maxSequentiallyHandledItems > 0 &&
-                        sequentiallyHandledItems >= maxSequentiallyHandledItems)
-                    {
-                        onYield();
-                        return;
-                    }
-
                     item = queue.poll();
 
                     if (item == null)
@@ -189,11 +189,7 @@ public final class AsyncQueueHandler<T>
     public void cancel()
     {
         running.set(false);
-
-        synchronized (syncRoot)
-        {
-            cancel(true);
-        }
+        cancel(true);
     }
 
     /**
@@ -219,10 +215,17 @@ public final class AsyncQueueHandler<T>
      */
     private void onYield()
     {
-        logger.fine("Yielding AsyncQueueHandler associated with "
-            + "AsyncQueueHandler with ID = " + id);
-        cancel(false);
-        readerFuture = executor.submit(reader);
+        if (logger.isLoggable(java.util.logging.Level.FINE))
+        {
+            logger.fine("Yielding AsyncQueueHandler associated with "
+                + "AsyncQueueHandler with ID = " + id);
+        }
+
+        synchronized (syncRoot)
+        {
+            cancel(false);
+            readerFuture = executor.submit(reader);
+        }
     }
 
     /**
@@ -233,10 +236,13 @@ public final class AsyncQueueHandler<T>
      */
     private void cancel(boolean mayInterruptIfRunning)
     {
-        if (readerFuture != null)
+        synchronized (syncRoot)
         {
-            readerFuture.cancel(mayInterruptIfRunning);
-            readerFuture = null;
+            if (readerFuture != null)
+            {
+                readerFuture.cancel(mayInterruptIfRunning);
+                readerFuture = null;
+            }
         }
     }
 
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -112,14 +112,14 @@ public final class AsyncQueueHandler<T>
 
             while (running.get())
             {
-                T item;
-
                 if (maxSequentiallyHandledItems > 0 &&
                     sequentiallyHandledItems >= maxSequentiallyHandledItems)
                 {
                     onYield();
                     return;
                 }
+
+                T item;
 
                 synchronized (syncRoot)
                 {
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright @ 2018 Jitsi.org
+ * Copyright @ 2018 8x8, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -235,6 +235,16 @@ public final class AsyncQueueHandler<T>
         String id,
         ExecutorService executor)
     {
+        if (queue == null)
+        {
+            throw new IllegalArgumentException("queue is null");
+        }
+
+        if (handler == null)
+        {
+            throw new IllegalArgumentException("handler is null");
+        }
+
         this.executor = executor != null ? executor : sharedExecutor;
         this.queue = queue;
         this.handler = handler;
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -211,14 +211,12 @@ public final class AsyncQueueHandler<T>
     /**
      * Invoked when execution of {@link #reader} is about to temporary
      * cancel and further execution need to be re-scheduled.
-     * Assuming called when lock on {@link #syncRoot} is already taken.
      */
     private void onYield()
     {
         if (logger.isLoggable(java.util.logging.Level.FINE))
         {
-            logger.fine("Yielding AsyncQueueHandler associated with "
-                + "AsyncQueueHandler with ID = " + id);
+            logger.fine("Yielding AsyncQueueHandler with ID = " + id);
         }
 
         synchronized (syncRoot)
@@ -229,8 +227,7 @@ public final class AsyncQueueHandler<T>
     }
 
     /**
-     * Attempts to cancel currently running reader. Assuming called when
-     * lock on {@link #syncRoot} is already taken
+     * Attempts to cancel currently running reader.
      * @param mayInterruptIfRunning indicates if {@link #reader} allowed
      * to be interrupted if running
      */
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -240,7 +240,6 @@ public final class AsyncQueueHandler<T>
 
         synchronized (syncRoot)
         {
-            cancel(false);
             readerFuture = executor.submit(reader);
         }
     }
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -207,7 +207,6 @@ public final class AsyncQueueHandler<T>
      */
     public void cancel()
     {
-        running.set(false);
         cancel(true);
     }
 
@@ -221,8 +220,7 @@ public final class AsyncQueueHandler<T>
         {
             if (readerFuture == null || readerFuture.isDone())
             {
-                running.set(true);
-                readerFuture = executor.submit(reader);
+                rescheduleReader();
             }
         }
     }
@@ -238,10 +236,7 @@ public final class AsyncQueueHandler<T>
             logger.fine("Yielding AsyncQueueHandler with ID = " + id);
         }
 
-        synchronized (syncRoot)
-        {
-            readerFuture = executor.submit(reader);
-        }
+        rescheduleReader();
     }
 
     /**
@@ -253,11 +248,25 @@ public final class AsyncQueueHandler<T>
     {
         synchronized (syncRoot)
         {
+            running.set(false);
+
             if (readerFuture != null)
             {
                 readerFuture.cancel(mayInterruptIfRunning);
                 readerFuture = null;
             }
+        }
+    }
+
+    /**
+     * Reschedules execution of {@link #reader} on {@link #executor}'s thread.
+     */
+    private void rescheduleReader()
+    {
+        synchronized (syncRoot)
+        {
+            running.set(true);
+            readerFuture = executor.submit(reader);
         }
     }
 
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -64,13 +64,14 @@ public final class AsyncQueueHandler<T>
     private final static long emptyQueueTimeoutNanoseconds = 50;
 
     /**
-     * Executor service to run <tt>AsyncQueueHandler</tt>, which asynchronously
+     * Executor service to run {@link #reader}, which asynchronously
      * invokes specified {@link #handler} on queued items.
      */
     private final ExecutorService executor;
 
     /**
-     * An {@link BlockingQueue <T>} which will be read on separate thread.
+     * An {@link BlockingQueue <T>} whose items read on separate thread and
+     * processed by provided {@link #handler}.
      */
     private final BlockingQueue<T> queue;
 
@@ -133,7 +134,7 @@ public final class AsyncQueueHandler<T>
 
             while (!cancelled.get())
             {
-                T pkt;
+                T item;
 
                 synchronized (syncRoot)
                 {
@@ -171,15 +172,15 @@ public final class AsyncQueueHandler<T>
                         // up to hundred nanoseconds, compared to not
                         // timeout at all has about 5-10% better
                         // performance in micro-benchmark scenarios.
-                        pkt = queue.poll(
+                        item = queue.poll(
                             emptyQueueTimeoutNanoseconds, TimeUnit.NANOSECONDS);
                     }
                     catch (InterruptedException e)
                     {
-                        pkt = null;
+                        item = null;
                     }
 
-                    if (pkt == null)
+                    if (item == null)
                     {
                         cancel(false);
                         return;
@@ -191,7 +192,7 @@ public final class AsyncQueueHandler<T>
 
                 try
                 {
-                    handler.handleItem(pkt);
+                    handler.handleItem(item);
                 }
                 catch (Throwable e)
                 {
@@ -222,9 +223,9 @@ public final class AsyncQueueHandler<T>
      * Constucts instance of {@link AsyncQueueHandler<T>} which is capable of
      * asyncronous reading provided queue from thread borrowed from executor to
      * process items with provided handler.
-     * @param queue host queue which holds items to process
+     * @param queue thread-safe queue which holds items to process
      * @param handler an implementation of handler routine which will be
-     *                invoked per each item placed in the queue.
+     * invoked per each item placed in the queue.
      * @param id optional identifier of current handler for debug purpose
      * @param executor optional executor service to borrow threads from
      */
--- a/src/main/java/org/ice4j/util/AsyncQueueHandler.java
+++ b/src/main/java/org/ice4j/util/AsyncQueueHandler.java
@@ -1,0 +1,448 @@
+/*
+ * Copyright @ 2018 Jitsi.org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ice4j.util;
+
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Logger;
+
+/**
+ * Asynchronously reads items from provided {@link #queue} on separate thread
+ * borrowed from {@link #executor} and process items with specified handler.
+ * Thread is not blocked when queue is empty and returned back to
+ * {@link #executor} pool. New or existing thread is borrowed from
+ * {@link #executor} when queue is non empty and {@link #reader} is not
+ * running
+ *
+ * @author Yura Yaroshevich
+ */
+public final class AsyncQueueHandler<T>
+{
+    /**
+     * The {@link java.util.logging.Logger} used by the
+     * {@link AsyncQueueHandler} class and its instances for logging output.
+     */
+    private static final java.util.logging.Logger logger
+        = Logger.getLogger(AsyncQueueHandler.class.getName());
+
+    /**
+     * ScheduledExecutorService to implement delay during throttling in
+     * <tt>AsyncQueueHandler</tt>
+     */
+    private final static ScheduledExecutorService delayTimer
+        = Executors.newSingleThreadScheduledExecutor(
+            new CustomizableThreadFactory(
+                AsyncQueueHandler.class.getName() + "-timer-", true));
+
+    /**
+     * The default <tt>ExecutorService</tt> to run <tt>AsyncQueueHandler</tt>
+     * when there is no executor injected in constructor.
+     */
+    private final static ExecutorService sharedExecutor
+        = Executors.newCachedThreadPool(
+            new CustomizableThreadFactory(
+                AsyncQueueHandler.class.getName() + "-executor-", true));
+
+    /**
+     * Number of nanoseconds to consider queue is empty to exit reading loop in
+     * borrowed thread.
+     */
+    private final static long emptyQueueTimeoutNanoseconds = 50;
+
+    /**
+     * Executor service to run <tt>AsyncQueueHandler</tt>, which asynchronously
+     * invokes specified {@link #handler} on queued items.
+     */
+    private final ExecutorService executor;
+
+    /**
+     * An {@link BlockingQueue <T>} which will be read on separate thread.
+     */
+    private final BlockingQueue<T> queue;
+
+    /**
+     * The {@link Handler<T>} used to handle items read from
+     * {@link #queue} by {@link #reader}.
+     */
+    private final Handler<T> handler;
+
+    /**
+     * An identifier of current reader which is used for debugging purpose.
+     */
+    private final String id;
+
+    /**
+     * A flag which indicates if reading of {@link #queue} should be
+     * cancelled.
+     */
+    private final AtomicBoolean cancelled = new AtomicBoolean();
+
+    /**
+     * Synchronization object of current instance state, in particular
+     * used to resolve races between {@link #handleQueueItemsUntilEmpty()}
+     * and {@link #reader} exit. In particular synchronization object used to
+     * access to field {@link #readerFuture} and {@link #delayedFuture}.
+     */
+    private final Object syncRoot = new Object();
+
+    /**
+     * Throttle calculator which is used for counting handled items and
+     * computing necessary delay before processing next item when
+     * throttling is enabled.
+     */
+    private final ThrottleCalculator<T> throttler;
+
+    /**
+     * Stores <tt>Future</tt> of currently executing {@link #reader}
+     */
+    private Future<?> readerFuture;
+
+    /**
+     * Stores <tt>ScheduledFuture</tt> of currently delayed
+     * execution {@link #reader}
+     */
+    private ScheduledFuture<?> delayedFuture;
+
+    /**
+     * Perpetually reads item from {@link #queue} and uses
+     * {@link #handler} on each of them.
+     */
+    private final Runnable reader = new Runnable()
+    {
+        @Override
+        public void run()
+        {
+            final long maxSequentiallyHandledItems =
+                handler.maxSequentiallyHandledItems();
+
+            int sequentiallyHandledItems = 0;
+
+            while (!cancelled.get())
+            {
+                T pkt;
+
+                synchronized (syncRoot)
+                {
+                    long delay = throttler.getDelayNanos();
+                    if (delay > 0)
+                    {
+                        onDelay(delay, TimeUnit.NANOSECONDS);
+                        return;
+                    }
+
+                    if (maxSequentiallyHandledItems > 0 &&
+                        sequentiallyHandledItems >= maxSequentiallyHandledItems)
+                    {
+                        /*
+                        All instances of AsyncQueueHandler executed on
+                        single shared instance of ExecutorService to better
+                        use existing threads and to reduce number of idle
+                        threads when reader's queue is empty.
+
+                        Having limited number of threads to execute might
+                        lead to other problem, when queue is always
+                        non-empty reader will keep running, while other
+                        readers might suffer from execution starvation. One
+                        way to solve this, is to to artificially interrupt
+                        current reader execution and pump it via internal
+                        executor's queue.
+                        */
+                        onYield();
+                        return;
+                    }
+
+                    try
+                    {
+                        // It is observed that giving very small poll timeout,
+                        // up to hundred nanoseconds, compared to not
+                        // timeout at all has about 5-10% better
+                        // performance in micro-benchmark scenarios.
+                        pkt = queue.poll(
+                            emptyQueueTimeoutNanoseconds, TimeUnit.NANOSECONDS);
+                    }
+                    catch (InterruptedException e)
+                    {
+                        pkt = null;
+                    }
+
+                    if (pkt == null)
+                    {
+                        cancel(false);
+                        return;
+                    }
+                }
+
+                sequentiallyHandledItems++;
+                throttler.notifyItemHandled();
+
+                try
+                {
+                    handler.handleItem(pkt);
+                }
+                catch (Throwable e)
+                {
+                    logger.warning("Failed to handle item: " + e);
+                }
+            }
+        }
+    };
+
+    /**
+     * Runnable which is scheduled with delay to perform actual
+     * scheduling of {@link #reader} execution.
+     */
+    private final Runnable delayedSchedule = new Runnable()
+    {
+        @Override
+        public void run()
+        {
+            synchronized (syncRoot)
+            {
+                delayedFuture = null;
+            }
+            handleQueueItemsUntilEmpty();
+        }
+    };
+
+    /**
+     * Constucts instance of {@link AsyncQueueHandler<T>} which is capable of
+     * asyncronous reading provided queue from thread borrowed from executor to
+     * process items with provided handler.
+     * @param queue host queue which holds items to process
+     * @param handler an implementation of handler routine which will be
+     *                invoked per each item placed in the queue.
+     * @param id optional identifier of current handler for debug purpose
+     * @param executor optional executor service to borrow threads from
+     */
+    public AsyncQueueHandler(
+        BlockingQueue<T> queue,
+        Handler<T> handler,
+        String id,
+        ExecutorService executor)
+    {
+        this.executor = executor != null ? executor : sharedExecutor;
+        this.queue = queue;
+        this.handler = handler;
+        this.id = id;
+        this.throttler = new ThrottleCalculator<>(handler);
+    }
+
+    /**
+     * Attempts to stop execution of {@link #reader} if running
+     */
+    public void cancel()
+    {
+        cancelled.set(true);
+
+        synchronized (syncRoot)
+        {
+            cancel(true);
+        }
+    }
+
+    /**
+     * Checks if {@link #reader} is running on one of {@link #executor}
+     * thread and if no submits execution of {@link #reader} on executor.
+     */
+    public void handleQueueItemsUntilEmpty()
+    {
+        if (cancelled.get())
+        {
+            return;
+        }
+
+        if (handler == null)
+        {
+            logger.warning("No handler set, the reading will not start.");
+            return;
+        }
+
+        synchronized (syncRoot)
+        {
+            if ((readerFuture == null || readerFuture.isDone())
+                && (delayedFuture == null || delayedFuture.isDone()))
+            {
+                readerFuture = executor.submit(reader);
+            }
+        }
+    }
+
+    /**
+     * Invoked when execution of {@link #reader} is about to temporary
+     * cancel and further execution need to be re-scheduled.
+     * Assuming called when lock on {@link #syncRoot} is already taken.
+     */
+    private void onYield()
+    {
+        logger.fine("Yielding AsyncQueueHandler associated with "
+            + "AsyncQueueHandler with ID = " + id);
+        cancel(false);
+        readerFuture = executor.submit(reader);
+    }
+
+    /**
+     * Invoked when next execution of {@link #reader} should be delayed.
+     * Assuming called when lock on {@link #syncRoot} is already taken.
+     * @param delay the time from now to delay execution
+     * @param unit the time unit of the delay parameter
+     */
+    private void onDelay(long delay, TimeUnit unit)
+    {
+        logger.fine("Delaying AsyncQueueHandler associated with "
+            + "AsyncQueueHandler with ID = " + id + " for "
+            + unit.toNanos(delay) + "us");
+        cancel(false);
+        delayedFuture = delayTimer.schedule(delayedSchedule, delay, unit);
+    }
+
+    /**
+     * Attempts to cancel currently running reader. Assuming called when
+     * lock on {@link #syncRoot} is already taken
+     * @param mayInterruptIfRunning indicates if {@link #reader} allowed
+     * to be interrupted if running
+     */
+    private void cancel(boolean mayInterruptIfRunning)
+    {
+        if (delayedFuture != null)
+        {
+            delayedFuture.cancel(mayInterruptIfRunning);
+            delayedFuture = null;
+        }
+
+        if (readerFuture != null)
+        {
+            readerFuture.cancel(mayInterruptIfRunning);
+            readerFuture = null;
+        }
+    }
+
+    /**
+     * A simple interface to handle enqueued {@link T} items.
+     * @param <T> the type of the item.
+     */
+    public interface Handler<T>
+    {
+        /**
+         * Does something with an item.
+         * @param item the item to do something with.
+         */
+        void handleItem(T item);
+
+        /**
+         * Specifies max number {@link #handleItem(T)} invocation
+         * per {@link #perNanos()}
+         * @return positive number of allowed handled items in case of
+         * throttling must be enabled.
+         */
+        default long maxHandledItems() {
+            return -1;
+        }
+
+        /**
+         * Specifies time interval in nanoseconds for {@link #maxHandledItems()}
+         * @return positive nanoseconds count in case of throttling must be
+         * enabled.
+         */
+        default long perNanos() {
+            return -1;
+        }
+
+        /**
+         * Specifies the number of items allowed to be handled sequentially
+         * without yielding control to executor's thread. Specifying positive
+         * number will allow other possible queues sharing same
+         * {@link ExecutorService} to process their items.
+         * @return positive value to specify max number of sequentially handled
+         * items which allows implementation of cooperative multi-tasking
+         * between different {@link AsyncQueueHandler<T>} sharing
+         * same {@link ExecutorService}.
+         */
+        default long maxSequentiallyHandledItems()
+        {
+            return -1;
+        }
+    }
+
+    /**
+     * Helper class to calculate throttle delay when throttling is enabled.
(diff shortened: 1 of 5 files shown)


mstyura:
@bbaldino could you please review this PR when you have time 😃. Thank you very much in advance!

mstyura:
Adding performance measurements comparison with my previous PRs.
See additional tables in [#154](https://github.com/jitsi/ice4j/pull/154) and [#155](https://github.com/jitsi/ice4j/pull/155).

As a baseline I've used previosly measured version with my not yet merged PR: [#154](https://github.com/jitsi/ice4j/pull/154), [#155](https://github.com/jitsi/ice4j/pull/155)

## `20 conferences` x `3 peers` - baseline
| Name | Value | Comment
|-|-|-|
|Computer:                          | 3 GHz Intel Core i7 mac mini, osx 10.14	| 
|JVB version:                       | 1.1-20180615.220358-69 |
|ice4j version:                     | 2.0.0-20181105.145933-16 | # with PR 154, 155
|org.ice4j.TERMINATION_DELAY, ms    | 604800000 |
|Conf per server:                   | 20 |
|Peers per conf:                    | 3 |
|Est. RTP streams:                  | 180 |
|CPU load:                          | 50-64% |
|Active threads:                    | 789 |
|Network IN:                        | ~2431 KBit/s | 
|Network OUT:                       | ~4878 KBit/s|
|Average RTT:                       | 3-28-ms |
|ICE connection stability:          | stable |


## `20 conferences` x `3 peers` - libjitsi #445 + ice4j #156 & #157
| Name | Value | Comment
|-|-|-|
|Computer:                          | 3 GHz Intel Core i7 mac mini, osx 10.14	| 
|JVB version:                       | 1.1-20181017.180330-93 | # even so I've changed version here, I've checked that actual performance gain was due to
|ice4j version:                     | 2.0.0-20181105.145933-16 | # with PR 154, 155
|org.ice4j.TERMINATION_DELAY, ms    | 604800000 |
|Conf per server:                   | 20 |
|Peers per conf:                    | 3 |
|Est. RTP streams:                  | 180 |
|CPU load:                          | 48-61% |
|Active threads:                    | 680 |
|Network IN:                        | ~2440 KBit/s | 
|Network OUT:                       | ~4887 KBit/s |
|Average RTT:                       | ~3-5 ms |
|ICE connection stability:          | stable |


## `30 conferences` x `3 peers` - baseline
| Name | Value | Comment
|-|-|-|
|Computer:                          | 3 GHz Intel Core i7 mac mini, osx 10.14	| 
|JVB version:                       | 1.1-20180615.220358-69 |
|ice4j version:                     | 2.0.0-20181105.145933-16 | # with PR 154, 155
|org.ice4j.TERMINATION_DELAY, ms    | 604800000 |
|Conf per server:                   | 20 |
|Peers per conf:                    | 3 |
|Est. RTP streams:                  | 180 |
|CPU load:                          | 99-100% |
|Active threads:                    | 1177 |
|Network IN:                        | ~2659 KBit/s | 
|Network OUT:                       | ~5346 KBit/s|
|Average RTT:                       | ~946-1040 ms |
|ICE connection stability:          | stable |


## `30 conferences` x `3 peers` - libjitsi #445 + ice4j #156 & #157
| Name | Value | Comment
|-|-|-|
|Computer:                          | 3 GHz Intel Core i7 mac mini, osx 10.14	| 
|JVB version:                       | 1.1-20180615.220358-69 |
|libjitsi                           | 1.0-20180904.192703-363 | # plus PR 445
|ice4j version:                     | 2.0.0-20181105.145933-16 | # with PR 154, 155
|org.ice4j.TERMINATION_DELAY, ms    | 604800000 |
|Conf per server:                   | 20 |
|Peers per conf:                    | 3 |
|Est. RTP streams:                  | 180 |
|CPU load:                          | 85-98% | # mostly around 89 % 
|Active threads:                    | 999 |
|Network IN:                        | ~3464 KBit/s | # please note gain here
|Network OUT:                       | ~6903 KBit/s| # and here, looks like earlier it was unable to handle such load properly
|Average RTT:                       | ~10-1400 ms | # mostly stay low values like ~ 10 ms
|ICE connection stability:          | stable |

mstyura:
Started work on moving `RTPConnectorOutputStream` into `PacketQueue`, added `releasePacket` as opposite to `createPacket` in last commit to simplify pooling implementation.

mstyura:
Jenkins, test this please

mstyura:
Jenkins, test this please

bbaldino:
@mstyura apologies about last week, it ended up getting taken over by orientation stuff.  At this point I think I've lost track of everything you've got in flight here, could you kinda summarize things for me?  I think you responded to my question about test harnesses too but I can't remember which PR it was.  Maybe starting a discussion in the jitsi community would be the best place.

mstyura:
> At this point I think I've lost track of everything you've got in flight here, could you kinda summarize things for me? 

@bbaldino ok, I'll try to summarize:
There are several pending merge request from me: #154, #155, #156, #157, jitsi/libjitsi#445 which I've proposed to make single machine handle more audio-only conferences in my environment.

The state of PRs is the following:
* #154 - it looks like it was reviewed by you, because I see questions in discussion, but there was not much comments about code itself. So it think you are still going to look to it and ask to make some changes before merge.

* #155  - there was some questions regarding change which I've addressed - like pooling and threads count. Not sure if you have looked at latest revision. 

* #156 - there was some questions regarding backward compatibility, since it's not very important PR for me as compared to all other I'll fix it later, for now postponing it by marking WIP

* #157 - looks like no one reviewed this PR yet because lack of comments, but it's one of the strongest: with unit tests, micro-benchmark and it might be a good base for other PRs which will move additional classes to PacketQueue.

* jitsi/libjitsi#445 - looks like no one reviewed it yet.

>  I think you responded to my question about test harnesses too but I can't remember which PR it was.
 
It was in [#155](https://github.com/jitsi/ice4j/pull/155#issuecomment-438638439)

> Maybe starting a discussion in the jitsi community would be the best place.

let's see how it will go when new question will arise if any. Since both of us are online now and I'll try to be online for next 3 hours maybe we could proceed quicker with comments here on GitHub or some IM app (skype, whatsapp, whatever quicker than comments)

bbaldino:
Hey Yury, thanks.  The mini benchmark here looks like a good start, though I'm confused at all the results...which is the one you ended up implementing?  I assume `PacketQueueOld` is the original code?  I also think we'd like to see something that tests at a higher level...i.e. actually runs an ice agent and a client and does some sort of throughput test.  We'd want to test both a single connection, but also multiple (since a lot of these changes are re-using a single thread across many connections, so there may be regressions there).

Also, I'm trying to get through these but please (continue to) be patient, as they're quite dense and I don't think I'm as fresh with the ice4j code as you are at this point :) Thanks again for all the work.

mstyura:
> which is the one you ended up implementing?

@bbaldino I've ended up implementing `PacketQueue` class which has configurable - it is possible to use it with different `ExecutorService` implementations. The micro-benchmarks are just an example of different performance when `PacketQueue` is configured with different `ExecutorService`.

>  I assume PacketQueueOld is the original code?

Yes

>  I also think we'd like to see something that tests at a higher level...i.e. actually runs an ice agent and a client and does some sort of throughput test.

It is not possible because, currently this class is not used in `ice4j`. It is only used in [`SctpConnection`](https://github.com/jitsi/jitsi-videobridge/blob/56d941ae8634a2f2d781f00a7d1a259ee7d1d9d6/src/main/java/org/jitsi/videobridge/SctpConnection.java#L255) in `videobridge` via [`RawPacketQueue`](https://github.com/jitsi/libjitsi/blob/6c0a31d6193dcd2b012ae3ced656ad403403d9ff/src/org/jitsi/util/RawPacketQueue.java#L21) defined in `libjitsi`.

>  We'd want to test both a single connection, but also multiple (since a lot of these changes are re-using a single thread across many connections, so there may be regressions there).

Even with single user of this class currently to not introduce any regression caused by limiting number of threads I've chosen `ExecutorService newCachedThreadPool()` as a default `ExecutorService` implementation.

> Also, I'm trying to get through these but please (continue to) be patient, as they're quite dense and I don't think I'm as fresh with the ice4j code as you are at this point :) 

Please consider all PR as independent one by one, maybe it would be easier to handle and not to be overloaded with my changes.

mstyura:
Jenkins, test this please

mstyura:
Good day, @bbaldino, could you please have a look at this PR?

Here is performance comparison with latest `ice4j` master:

## before
| Name | Value | Comment
|-|-|-|
|VM CPU                             | 2 cores Intel(R) Xeon(R) CPU E5-2630 v2 @ 2.60GHz	| 
|RAM                                | 4 Gb |
|OS                                 | CentOS Linux release 7.5.1804 (Core)
|JVB version:                       | 1.1-20180615.220358-69 | # old version, i know
|libjitsi version:                  | 1.0-20180607.160234-352 |
|ice4j version:                     | 2.0.0-20181129.205550-18 | # latest master
|org.ice4j.TERMINATION_DELAY, ms    | 604800000 |
|Conf per server:                   | 25 |
|Peers per conf:                    | 5 |
|Est. RTP streams:                  | 625 |
|CPU load:                          | 89 - 95 % |
|Active threads:                    | 1406 |
|RTP loss rate download             | 0% |
|RTP loss rate upload               | < 2.7 % |
|Network IN:                        | ~ 4866 KBit/s | 
|Network OUT:                       | ~ 19503 KBit/s |
|Average RTT:                       | 5-15 ms |
|ICE connection stability:          | stable |

## after
| Name | Value | Comment
|-|-|-|
|VM CPU                             | 2 cores Intel(R) Xeon(R) CPU E5-2630 v2 @ 2.60GHz	| 
|RAM                                | 4 Gb |
|OS                                 | CentOS Linux release 7.5.1804 (Core)
|JVB version:                       | 1.1-20180615.220358-69 | # old version, i know
|libjitsi version:                  | 1.0-20180607.160234-352 |
|ice4j version:                     | 2.0.0-20181129.205550-18 | # + PR 157
|org.ice4j.TERMINATION_DELAY, ms    | 604800000 |
|Conf per server:                   | 25 |
|Peers per conf:                    | 5 |
|Est. RTP streams:                  | 625 |
|CPU load:                          | 90-94 % |
|Active threads:                    | 1288 | # org.jitsi.util.RawPacketQueue-SctpConnection-* were gone
|RTP loss rate download             | 0% |
|RTP loss rate upload               | < 1.7% | # get lower for some reason
|Network IN:                        | ~ 4903 KBit/s | 
|Network OUT:                       | ~ 19680 KBit/s |
|Average RTT:                       | 4 - 11 ms | # get lower for some reason
|ICE connection stability:          | stable |


BTW, I had a brief look into `nlj`, seems like very similar problem were solved, but slightly differently in [`RtpreceiverImpl.kt`](https://github.com/bbaldino/nlj/blob/8de285d6a5aea913ae4e730e2b66c8bab1c17582/src/main/kotlin/org/jitsi/nlj/RtpReceiverImpl.kt#L221) and in [`RtpSenderImpl.kt`](https://github.com/bbaldino/nlj/blob/8de285d6a5aea913ae4e730e2b66c8bab1c17582/src/main/kotlin/org/jitsi/nlj/RtpSenderImpl.kt#L154) 😄 

I could probably "backport" this PR as a generic class for `nlj` or (maybe even better) split current `PacketQueue` implementation into queue and something like `AsyncMessageHandler<TMessage>`, so `nlj` could use generic queue handler as well, instead of implementing it's own.

Anyway I believe that for now the change is still quite useful for `ice4j` as is.

Thank  you very much, Brian!

mstyura:
Jenkins, test this please

src/main/java/org/ice4j/util/PacketQueue.java
```diff
@@ -494,5 +501,332 @@ protected abstract T createPacket(
...
+                            readers might suffer from execution starvation. One
+                            way to solve this, is to to artificially interrupt
+                            current reader execution and pump it via internal
+                            executor's queue.
+                            */
+                            onYield();
+                            return;
+                        }
+
+                        pkt = queue.poll();
```
bbaldino: can you elaborate a bit on the scheme here?  i'm concerned because i tried a similar scheme in the bridge rewrite (polling queues so as to be able to shared threads) and i found it to be unusable due to the extra CPU involved in polling as opposed to just blocking on queues until there was data (with dedicated threads).  even after i played with reading them less frequently at some interval, there was always a CPU tradeoff in comparison to blocking until there was data.

bbaldino: (I will say I am trying a version of poll with a 100ms timeout, but this is just to allow an opportunity for the thread to shutdown cleanly.  I may investigate a 'poison' value on the queue to achieve this and avoid the poll altogether).

mstyura: @bbaldino I have a version of this with poison value as well, it did’t push it because I though I have better chances merging simpler version :)

bbaldino: but the poison still won't help you with sharing threads, no?

mstyura: > can you elaborate a bit on the scheme here? i'm concerned because i tried a similar scheme in the bridge rewrite (polling queues so as to be able to shared threads) and i found it to be unusable due to the extra CPU involved in polling as opposed to just blocking on queues until there was data (with dedicated threads). even after i played with reading them less frequently at some interval, there was always a CPU tradeoff in comparison to blocking until there was data.

I’ve just re-read this commment as well as kotlin implementation inside new libjitsi. As I can see, in a kotlin implementation there is always reschedule to poll, even if queue is currently empty. It looks like there is basically an infinite loop via recursion via scheduler queue and it will eat CPU to poll.
In my implementation I stop pollling and release thread when there is no lot items to proceed. The trigger to start polling again is a queue’s “add” method. So there must be no extra work to poll in both cases: when queue is empty and not empty, and even when adding and removing to queue is done at the same rate.

bbaldino: The code pushed to github may be a bit stale, but that's a good idea (not triggering another read until the next add)!  I will experiment with that!  (Also, the call to schedule is recursive but since executor.execute just schedules the job we don't actually get an infinite recursive stack).

mstyura: > Also, the call to schedule is recursive but since executor.execute just schedules the job we don't actually get an infinite recursive stack


Yes. I got that trick :)


>  I will experiment with that!


You may try experimenting with this PR, there are variety of examples of different configuration of packet queue with different executor service, throttling and cooperative multi-tasking configuration inside benchmark tests. You could run them easily for comparison as well as introduce yours :)

mstyura:
Good day, @bbaldino!

I've verified what you've asked: the proposed implementation does not have a problem of eating CPU for polling when `PacketQueue` does not have work. Thread borrowing from `ExecutorService` confirmed to be triggered only when item is added to queue. I've also added unit test to check that `ExecutorService` thread is not used, when `PacketQueue` completed handling it's items. Test confirms that millions of `PacketQueue` instances can share single thread.
Could you please review PR and see if anything else need to be fixed before getting your approval?

Thank you very much, Brian!

mstyura:
Good day, @bbaldino!

[You've asked me](https://github.com/jitsi/ice4j/pull/155#issuecomment-438464759) in one of the previous pull request if I have a working solution to do load testing [`videobridge`](https://github.com/jitsi/jitsi-videobridge). My answer at that time was "yes, but my load solution relies on my custom REST API around JVB". Even so I have a solution to load [JVB](https://github.com/jitsi/jitsi-videobridge) and could share it, it will not be very useful, because I can not share REST API wrapper part (and nobody need it except me). 
Recently there was [a new project appeared](https://github.com/Hartigan/jitsi-videobridge-test/) which is capable of load testing [JVB](https://github.com/jitsi/jitsi-videobridge) with audio-only conferences via standard colibri-based REST API. This project is very-very similar to what I use for load-test `JVB` via my custom REST API, when I provide comparison tables in PRs. 
So, you might try using it and it might be eventually adopted by `jitsi` as a replacement of (AFAIK) not currently supported [jitsi-hammer](https://github.com/jitsi/jitsi-hammer).
I hope that ability of doing load testing over latest [videobridge](https://github.com/jitsi/jitsi-videobridge) with [above-mentioned tool](https://github.com/Hartigan/jitsi-videobridge-test) will have positive influence on speed of accepting different performance-related PR 😄 .
Hope you'll have a change to take a looks at this PR today 🤞 .

Thank you very much, Brian!

bbaldino:
Hey @mstyura, thanks for the link, that's very interesting.  Could it support video as well?  I don't know if I will have time to look at this today, as it's quite dense and I need some time to go through it and I'm still trying to focus on the new bridge work.

mstyura:
> Could it support video as well?

Right now video is not supported, but support could be added relatively easily.

bbaldino:
> > Could it support video as well?
> 
> Right now video is not supported, but support could be added relatively easily.

Ah, that's very exciting.  We've been using a selenium but it's reliability isn't the best.  I'll have to take a look at the project.

src/main/java/org/ice4j/util/PacketQueue.java
```diff
@@ -494,5 +501,332 @@ protected abstract T createPacket(
...
+
+                int sequentiallyHandledPackets = 0;
+
+                while (!closed.get())
+                {
+                    T pkt;
+
+                    synchronized (syncRoot)
+                    {
+                        long delay = throttler.getDelayNanos();
```
bbaldino: i'm only taking some quick looks here, but i don't quite get the throttler.  i thought the idea was:

read from the queue and if:
1) it was empty -> this context does nothing until it is rescheduled via something being added to the queue
2) there is data, and we're within max sequential packets -> process the data
3) there is data, and we're now beyond max sequential packets -> wait some delay before trying to read again

what case is the throttler handling that max sequential isn't?

mstyura: @bbaldino the notion of throttler was borrowed [from `libjitsi`](https://github.com/jitsi/libjitsi/blob/6c0a31d6193dcd2b012ae3ced656ad403403d9ff/src/org/jitsi/impl/neomedia/RTPConnectorOutputStream.java#L937). I've added it here because I'm going to replace class in `libjitsi` by this `PacketQueue` from `ice4j`.
`maxSequentiallyProcessedPackets` is for implementing cooperative multi-tasking: for example, when some queue  always has something to proceed and it shares executor thread with some other queue then it is important to switch execution between different queues, so all queues are processed and there is no queues which suffer from execution starvation. 
Throttling and `maxSequentiallyProcessedPackets` could be turned on/off independently of each other, so it's quite fine to do both checks in this while loop to achieve both throttling and cooperative multi tasking.

mstyura: My understand of throttling is following: queue should not proceed more than `N` items per `P` seconds. For example, it could be configured that there should be no more than 100 processed items per second, which is completely different from max items processed in row, which is controlled by `maxSequentiallyProcessedPackets `.

bbaldino: Ok, I understand.  I'm not sure if we really need that, and would be great to simplify this (I understand it can be disabled, but it does add complexity).  For what it's worth, the usage of this will be gone in the bridge rewrite, but I would love to be able to use this scheme on a LinkedBlockingQueue, as we discussed.  I'm not sure we necessarily need to block this work on the rewrite, but, maybe we can at least make AsyncPacketReader a first class citizen (a class in its own file) as I think it's a useful class.

mstyura:
Good day, @bbaldino!

I've extracted the queue reading logic into dedicated class `AsyncQueueHandler<T>` (former `AsyncPacketReader`) from `PacketQueue<T>`. 
I do like a lot this change which you proposed, because it made logic inside `AsyncQueueHandler<T>` reusable and it is as well significantly reduced diff inside `PacketQueue<T>` which made it simpler to review. Don't be scared by amount of added lines in diff, most of them comes from unit tests and benchmark. The rest of the code of `AsyncQueueHandler` you probably already seen earlier.
I still preserved throttling logic, because I absolutely need it in old `libjitsi`, until new one is ready and stable. I don't mind removing it later, it should easy to done.

Could you please review changes today? I'll be online until `13.00 PST` and could quickly react on questions and other fixes you might ask to implement.

Thank you very much, Brian!

mstyura:
Jenkins test this please

bbaldino:
hey @mstyura i am going to try and look at these today.  i planned to first thing this morning but another issue has come up that needs to be looked at first.  hopefully i'll be able to look at this today, though.

yes i'd like to be able to use the AsyncReader logic in the new bridge as a good compromise between non-blocking threads and not constantly polling and therefore using too much cpu, so i'd like to be able to get it in so i can use it.

bbaldino:
Don't worry about those test failures, there is a known issue with the torture tests right now that's being worked on.

mstyura:
If there are some necessary changes required before accepting this PR feel free to add fixing commits to this PR, because I'm likely to be not available later today.

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,459 @@
...
+         */
+        private long intervalStartTimeNanos = 0;
+
+        /**
+         * {@link Handler <T>} instance which provide throttling
+         * configuration
+         */
+        private final Handler<T> handler;
+
+        ThrottleCalculator(Handler<T> handler)
```
bbaldino: i'm still getting stuck on the need for the throttling.  i disagree about the amount of complexity it adds, because although the mechanism itself may be somewhat simple there's an extra future that needs to be tracked and synchronized correctly, as well as logic that needs to be understood and properly dealt with if any changes are to be made (i took a look at removing the logic for this, the file has over 25% fewer lines of code with it gone). since it is not being used i'm feeling like we should just remove this.  to me, it even feels unlikely to use the maxSequential field.

i know you are just porting over what's there, just my thoughts.

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,459 @@
...
+     */
+    private final AtomicBoolean cancelled = new AtomicBoolean();
+
+    /**
+     * Synchronization object of current instance state, in particular
+     * used to resolve races between {@link #handleQueueItemsUntilEmpty()}
+     * and {@link #reader} exit. In particular synchronization object used to
+     * access to field {@link #readerFuture} and {@link #delayedFuture}.
+     */
+    private final Object syncRoot = new Object();
```
bbaldino: i find the usage of this hard to follow.  you mention here it protects `readerFuture`, but it appears you synchronize on this across a much larger scope in multiple places.  (btw I looked through this after already removing the throttler in my local checkout, so these comments center around the sync'ing of `readerFuture`)

1) in `cancel()` you synchronize on it before calling `cancel(boolean)`.  `cancel(boolean)` does write to readerFuture, but why not just synchronize on `syncRoot` in `cancel(boolean)` directly?  I see you denote that that method assumes the `syncRoot` has already been locked, but why not just do it there?  it'd make it harder for others to make mistakes. 

2) The use in `handleQueueItemsUntilEmpty` seems fine

3) The block in the Runnable's `run` is very big and protects all sorts of things.  As far as I can tell it's only needed in `onYield` and `cancel` (which would be taken care of by #1 above).  I thought perhaps making the lock tighter here would result in acquiring it more than once (in both `onYield` and `cancel`), but we return after calling `onYield` so I don't think that's the case.  EDIT: actually this is wrong.  in `onYield` we call cancel, so we'll lock twice there, but given this case is likely rare it seems a reasonable tradeoff?  My thinking is to try and minimize the likelihood someone could mess up the synchronization in the future.

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,459 @@
...
+        {
+            cancel(true);
+        }
+    }
+
+    /**
+     * Checks if {@link #reader} is running on one of {@link #executor}
+     * thread and if no submits execution of {@link #reader} on executor.
+     */
+    public void handleQueueItemsUntilEmpty()
```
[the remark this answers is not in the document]
mstyura: @bbaldino I had the following example when choosen the name of method, let's consider there is a queue with `5` items and `maxSequentiallyHandledItems = 1`, so whenever single item is processed there is a process of releasing and re-acquiring execution thread from `ExecutorService` inside `AsyncQueueHandler`. So, when I calls `AsyncQueueHandler.maxSequentiallyHandledItems` on a queue which has 5 items I expect that `AsyncQueueHandler` will process all of the items until queue become empty, no matter that `maxSequentiallyHandledItems == 1`. The `maxSequentiallyHandledItems` only affects how `AsyncQueueHandler` under the hood implement thread sharing with other instances sharing same `ExecutorService`.
Could you please suggest the name if you still think I should rename it?

src/main/java/org/ice4j/util/AsyncQueueHandler.java
mstyura: ok, removing. Just curios what is the timeline of new bridge arriving, so I will not need this functionality?

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,459 @@
...
+    /**
+     * An identifier of current reader which is used for debugging purpose.
+     */
+    private final String id;
+
+    /**
+     * A flag which indicates if reading of {@link #queue} should be
+     * cancelled.
+     */
+    private final AtomicBoolean cancelled = new AtomicBoolean();
```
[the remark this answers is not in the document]
mstyura: Renamed running

src/main/java/org/ice4j/util/AsyncQueueHandler.java
mstyura: > 1. in cancel() you synchronize on it before calling cancel(boolean). cancel(boolean) does write to readerFuture, but why not just synchronize on syncRoot in cancel(boolean) directly? I see you denote that that method assumes the syncRoot has already been locked, but why not just do it there? it'd make it harder for others to make mistakes.

I've considered nested `synchronized` blocks as unnecessary, that's why I've initially added private methods which are assumed to be executed under synchronized section. I've checked with benchmark that having nested sections does not hurt performance, so I've added it to simplify reading as you've asked.

> 3. The block in the Runnable's run is very big and protects all sorts of things. As far as I can tell it's only needed in onYield and cancel (which would be taken care of by #1 above). I thought perhaps making the lock tighter here would result in acquiring it more than once (in both onYield and cancel), but we return after calling onYield so I don't think that's the case. EDIT: actually this is wrong. in onYield we call cancel, so we'll lock twice there, but given this case is likely rare it seems a reasonable tradeoff? My thinking is to try and minimize the likelihood someone could mess up the synchronization in the future.

I've removed throttling and moved yield out of synchronized block, but poll and cancel still must be under single synchronized block. The reason why poll & cancel must be under **single** synchronized section is that there is inevitable race between `queue.poll` which detects queue is empty and hence reader is going to exit (`cancel` called), and item being added just between `queue.poll` and `cancel` and hence reader must be rescheduled, but `handleQueueItemsUntilEmpty` count not observe that because `Future` is neither null or done at that moment.

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -127,58 +108,22 @@
...
-                        // timeout at all has about 5-10% better
-                        // performance in micro-benchmark scenarios.
-                        item = queue.poll(
-                            emptyQueueTimeoutNanoseconds, TimeUnit.NANOSECONDS);
-                    }
-                    catch (InterruptedException e)
-                    {
-                        item = null;
-                    }
+                    item = queue.poll();
```
bbaldino: based on your previous comment:

> // It is observed that giving very small poll timeout,	
>                         // up to hundred nanoseconds, compared to not	
>                         // timeout at all has about 5-10% better	
>                         // performance in micro-benchmark scenarios.

did you find that this was no longer the case?

src/main/java/org/ice4j/util/AsyncQueueHandler.java
bbaldino: > I've removed throttling and moved yield out of synchronized block, but poll and cancel still must be under single synchronized block. The reason why poll & cancel must be under single synchronized section is that there is inevitable race between queue.poll which detects queue is empty and hence reader is going to exit (cancel called), and item being added just between queue.poll and cancel and hence reader must be rescheduled, but handleQueueItemsUntilEmpty could not observe that because Future is neither null or done at that moment.

Good catch

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,261 @@
...
+                running.set(true);
+                readerFuture = executor.submit(reader);
+            }
+        }
+    }
+
+    /**
+     * Invoked when execution of {@link #reader} is about to temporary
+     * cancel and further execution need to be re-scheduled.
+     * Assuming called when lock on {@link #syncRoot} is already taken.
```
bbaldino: think this comment line no longer applies

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,261 @@
...
+     */
+    private void onYield()
+    {
+        if (logger.isLoggable(java.util.logging.Level.FINE))
+        {
+            logger.fine("Yielding AsyncQueueHandler associated with "
+                + "AsyncQueueHandler with ID = " + id);
```
bbaldino: should this log just be `"Yielding AsyncQueueHandler associated with ID " + id`?

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,261 @@
...
+
+    /**
+     * Attempts to cancel currently running reader. Assuming called when
+     * lock on {@link #syncRoot} is already taken
+     * @param mayInterruptIfRunning indicates if {@link #reader} allowed
+     * to be interrupted if running
+     */
+    private void cancel(boolean mayInterruptIfRunning)
+    {
+        synchronized (syncRoot)
```
bbaldino: i think the `running.set(false);` needs to be in this method, no?  some inner methods call this `cancel(boolean)` directly, and therefore won't change `running` since it's only set in `cancel()`?

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,261 @@
...
+     * @param maxSequentiallyHandledItems maximum number of items sequentially
+     * handled on thread borrowed from {@link #executor} before temporary
+     * releasing thread and re-acquiring it from {@link #executor}.
+     */
+    public AsyncQueueHandler(
+        BlockingQueue<T> queue,
+        Handler<T> handler,
+        String id,
+        ExecutorService executor,
+        long maxSequentiallyHandledItems)
```
bbaldino: nit: can we add another ctor that doesn't take in a `maxSequentiallyHandledItems` and defaults it to no limit?

src/main/java/org/ice4j/util/AsyncQueueHandler.java
mstyura: > maybe i misunderstood something, i didn't think this was actually being used anywhere (implemented, yes, but was the functionality actually in use?)

I personally didn't check if it's invoked in runtime at the moment of implementing it in `PacketQueue` and for my use case of audio only conferences it doesn't not happen, because the only place this invoked is from [VideoMediaStreamImpl](https://github.com/jitsi/libjitsi/blob/6c0a31d6193dcd2b012ae3ced656ad403403d9ff/src/org/jitsi/impl/neomedia/VideoMediaStreamImpl.java#L694).

But since you've asked, I've just checked conference with video stream under debugger and it looks like by default this code is not invoked, because `maxBandwith` is `2^31-1`.

src/main/java/org/ice4j/util/AsyncQueueHandler.java
mstyura: > did you find that this was no longer the case?

Honestly saying I was afraid that this trick will not survive review, so I've preemptively removed it to save one cycle of review latency 😄 and I've done that micro-benchmark on Windows, which might give different result than when running on Linux, so that was another reason to remove (I am going to add it back as separate request when I confirm it's benefit on Linux as well, because smaller PR is easier to handle)

bbaldino: Ah ok...yeah that sounds good (checking it and doing it in another PR)

src/main/java/org/ice4j/util/AsyncQueueHandler.java
mstyura: > i think the running.set(false); needs to be in this method, no? 

Actually no, because, for example `cancel(boolean)` is called from `onYield`, which just reschedules `reader`, so if I add `running.set(false)` to `cancel(boolean)` then `onYield()` must immediately undo this flag by calling `running.set(true)`...
and it feels like `running` is actually not what it's named, the actual reader may not physically run, but being in queue of corresponding `ExecutorService` waiting it's chance to execute.... So maybe it's better to rollback to negative name which exactly describes what that flag was actually about:  it had meaning that no processing and no further re-scheduling should be done because user of class asked to stop processing?

mstyura:
@bbaldino I think I'm done, the only unresolved question is about `running` flag in [comment](https://github.com/jitsi/ice4j/pull/157#discussion_r240739156), so waiting for your reply to it.

src/main/java/org/ice4j/util/AsyncQueueHandler.java
bbaldino: ah i see.  in my opinion it's not `running` that's misleading there.  actually, i wonder if we should extract the onYield code to some `rescheduleReader` private helper? like:
```
// 
private void rescheduleReader()
{
  synchronized (syncRoot)
  {
    if (readerFuture != null)
    {
      readerFuture.cancel(false);
    }
    readerFuture = executor.submit(reader);
  }
}
```

maybe `handleQueueItemsUntilEmpty` could also use this?

mstyura: Done this. It's occurred that after removing extra future used for throttling there no need to cancel existing future in `onYield`, it's possible just override the variable. I've added `rescheduleFuture` which is called from `onYield` and `handleQueueItemsUntilEmpty `. So, now loos better, IMO.

bbaldino:
thanks @mstyura, i think this looks good now.  I'm just trying to clarify something about the license that should be included in the new file.

src/main/java/org/ice4j/util/AsyncQueueHandler.java
```diff
@@ -0,0 +1,285 @@
+/*
+ * Copyright @ 2018 Jitsi.org
```
bbaldino: ok @mstyura last change: can you make this say `8x8, Inc.` instead of `Jitsi.org` here?

mstyura: Done :)

mstyura:
@bbaldino is there any known issue with `Jenkins builder` such that new version of `ice4j` is not in [jitsi/jitsi-maven-repository](https://github.com/jitsi/jitsi-maven-repository)?

bbaldino:
@mstyura we're transitioning some infrastructure right now, so the jenkins builders are paused.  we're hoping to have things back up tomorrow.

damencho:
I was enabling back ice4j build and I ran into this:
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /mnt/bigci/jobs/ice4j/workspace/ice4j/src/main/java/org/ice4j/stack/MessageProcessor.java:[89,66] cannot find symbol
  symbol:   method getMessageQueue()
  location: variable netAccessManager of type org.ice4j.stack.NetAccessManager
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE

It builds fine locally ... but I have a problem compiling it on the ci machine.

mstyura:
Hi, @damencho,
The log suggest that the compilation issue caused by #155, because I've removed this method.

damencho:
@mstyura but I still do not get it, why locally it is ok and on the ci machine it fails. Latests source and same java versions on both places ... and by not seeing what is a problem, I'm wondering how to fix ci so it can produce new build :P

mstyura:
@damencho sounds like `git checkout` went wrong on CI, because this file had to be removed and not included into compilation: https://github.com/jitsi/ice4j/tree/master/src/main/java/org/ice4j/stack - no such file here, but only `MessageProcessingTask.java`. Can you `ssh` on CI and check an output of `git status`?

damencho:
Good guess ... at some point we were doing some rsync which didn't delete files ... Thanks. Build is comming in maven ... :)

PR #164, opened 2018-12-17
Invoke `releasePacket` on dropped items in `PacketQueue`.

mstyura:
There was missing call to `releasePacket` when packet was dropped from queue due to queue overflow.

To facilitate usage of object pool in conjunction with `PacketQueue` the missing call to `releasePacket` were added.

A scope of `synchronized` block inside `doAdd` were reduced to just notify objects waiting on `queue`.  The underlying queue is thread-safe and does not need a protection with `syncronized` block. It is also allowed to avoid calling `releasePacket` within `synchronized` block.

A unit test releasing missing call to `releasePacket` were added.

### DIFF
--- a/src/main/java/org/ice4j/util/PacketQueue.java
+++ b/src/main/java/org/ice4j/util/PacketQueue.java
@@ -279,31 +279,36 @@ public abstract class PacketQueue<T>
         if (closed)
             return;
 
+        while (!queue.offer(pkt))
+        {
+            // Drop from the head of the queue.
+            T p = queue.poll();
+            if (p != null)
+            {
+                if (queueStatistics != null)
+                {
+                    queueStatistics.remove(System.currentTimeMillis());
+                }
+                if (logDroppedPacket(++numDroppedPackets))
+                {
+                    logger.warning(
+                        "Packets dropped (id=" + id + "): " + numDroppedPackets);
+                }
+
+                // Call release on automatically popped packet to allow
+                // proper implementation of object pooling by PacketQueue
+                // users
+                releasePacket(p);
+            }
+        }
+
+        if (queueStatistics != null)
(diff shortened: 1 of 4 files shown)


mstyura:
Jenkins test this please.

mstyura:
Jenkins test this please

mstyura:
Jenkins test this please

mstyura:
Good day, @bbaldino!

Could you please have a look at this PR which fixes minor issue inside `PacketQueue` - the `releasePacket` was not invoked for dropped packets which is not very suitable for queue users which employs packets pooling.

Thank you very much, Brian!

src/main/java/org/ice4j/util/PacketQueue.java
```diff
@@ -279,32 +280,39 @@ private void doAdd(T pkt)
...
-                    }
-                    if (logDroppedPacket(++numDroppedPackets))
-                    {
-                        logger.warning(
-                            "Packets dropped (id=" + id + "): " + numDroppedPackets);
-                    }
+                    queueStatistics.remove(System.currentTimeMillis());
+                }
+                final int numDroppedPackets =
+                    this.numDroppedPackets.incrementAndGet();
```
bbaldino: so it looks like this doesn't get accessed from anywhere else, is that right?  does this method get called from multiple threads?

mstyura: @bbaldino it might be, there is no protection against calling `PacketQueue.add` from multiple threads.

mstyura:
Thank you very much for review and approving this PR, @bbaldino! Could you please also hit merge button when tests on Jenkins suddenly become green again (looks like failure is unrelated to my changes)?

PR #165, opened 2018-12-18
Use notify instead of notifyAll when appropariate on hotpaths

mstyura:
Use `Object.notify` instead of `Object.notifyAll` when only one thread is affected by notification.
`notify` has a slightly better performance according to profiling done with `Visual VM`.


mstyura:
Good day, @bbaldino!

Could you please have a look at this very tiny PR?

Thank you very much!

PR #167, opened 2018-12-20
WIP: Fixed socket receive buffer interface misuse

mstyura:
Use type with tiny interface (offer/poll/scan/isEmpty) for socket receive buffer.


bbaldino:
Hey Yury, I'll be out until the 2nd, but feel free to assign PRs to me and
I can look into them when I get back.

src/main/java/org/ice4j/socket/SocketReceiveBuffer.java
```diff
@@ -91,102 +81,39 @@ public boolean isEmpty() {
      */
     public boolean offer(DatagramPacket p)
     {
-        boolean added = buffer.offer(p);
-
-        // Keep track of the (total) size in bytes of this receive buffer in
-        // order to be able to enforce SO_RCVBUF.
-        if (added)
+        while (!buffer.offer(p))
```
paweldomas: do I understand this loop correctly that it's going to drop packets until there's space available ? Doesn't the API say that it would just return false instead ?

paweldomas: Oh I see that in the following commit the method is renamed. Does it make sense to squash the commits ?

mstyura: @paweldomas yes, but usually they were squashed automatically by GitHub when “merge PR” button were hit. Do you want me squash them manually?


> do I understand this loop correctly that it's going to drop packets until there's space available ?


Yes. The oldest packets are dropped. This is what usually done in other places in ice4j and LibJitsi when such overflow is happen.


> Doesn't the API say that it would just return false instead ?
 

It was not handled by caller before and it still unclear what caller can do with that return value - so, I’ve desided to implement the drop behavior on buffer itself.

mstyura:
Jenkins test this please.

mstyura:
@bgrozev thank you very much for reviewing and approving this PR, Boris! Could you please also hit merge button? Thanks!

PR #168, opened 2018-12-27
Use pool of datagram packets inside MultiplexingXXXSocketSupport

mstyura:
**TL; DR**: Use pool of datagram packets inside MultiplexingXXXSocketSupport to reduce clone() overhead.

**Problem**:
According to sampling profiler (Oracle Visual VM) there is significant overhead caused by [`clone()`](https://github.com/jitsi/ice4j/blob/c56fcc3a3f86ccfda8ce57362d378fc11bd35824/src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java#L626)'ing `DatagramPackets` before receiving data from underlying socket:

**Solution**:
The actual receive from underlying socket is no longer invoked with a shallow clone of datagram packet `p` passed to [`MultiplexingXXXSocketSupport.receive(SocketReceiveBuffer received, DatagramPacket p, int timeout)`](https://github.com/jitsi/ice4j/blob/c56fcc3a3f86ccfda8ce57362d378fc11bd35824/src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java#L542), but with `DatagramPacket` taken from introduced pool.


src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
```diff
@@ -188,6 +188,24 @@ else if (logger.isLoggable(Level.WARNING))
...
+    /**
+     * The max number of {@link DatagramPacket}'s stored in the receive pool.
+     */
+    private static final int RECEIVE_DATAGRAM_PACKETS_POOL_SIZE = 64;
+
+    /**
+     * The buffer size of pooled datagram packet used to read from
+     * underlying socket.
+     */
+    private static final int POOLED_DATAGRAM_PACKET_SIZE = 1500;
```
paweldomas: Do you know how is this value related to the 'receiveBufferSize' ? do you think it would be a problem if the buffer size would be bigger ? I see you've chosen Ethernet's MTU as default which I guess should be fine most of the time, but what bothers me is that this socket allows to adjust the buffer's size.

mstyura: > Do you know how is this value related to the 'receiveBufferSize' ?

This constant limits the max datagram packet size read from Java’s DatagramSocket.
`receiveBufferSize` controls the size of kernel buffer for socket (SO_RECVBUF), but in this case the socket is kind of virtual and this value controls the max total number of bytes stored across all datagrams stored in receive buffer - which is in this case just List<DatagramPakcet>(until #167 merged). As I understand it - it just a programmatic simulation of kernel behavior which limits the number of bytes socket can have “unread”, before start dropping the most old packets.


> do you think it would be a problem if the buffer size would be bigger ?

If I understand your question correctly, the no, since the `receiveBufferSize` and `POOLED_DATAGRAM_PACKET_SIZE` controls different things.


It probably make sense to have such many packets in pool as `receiveBufferSize` allow in receive buffer, but until #167 merged the socket receive buffer  is just a `List<DatagramPacket>` which makes it inconvinient to add pooling directly inside socket receive buffer.

src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
```diff
@@ -648,9 +648,11 @@ void receive(List<DatagramPacket> received, DatagramPacket p, int timeout)
                 {
                     receivePacket = new DatagramPacket(
                         new byte[POOLED_DATAGRAM_PACKET_SIZE],
-                        0, POOLED_DATAGRAM_PACKET_SIZE);
+                        POOLED_DATAGRAM_PACKET_SIZE);
                 }
 
+                receivePacket.setLength(POOLED_DATAGRAM_PACKET_SIZE);
```
paweldomas: I wonder if that's really necessary. I've been reading this code and it looks like there's some confusion:
https://github.com/jitsi/ice4j/blob/master/src/main/java/org/ice4j/socket/DelegatingDatagramSocket.java#L682
How about we reseting both length and offset ?

mstyura: @paweldomas I though it is mandatory, because the [documentation says](https://docs.oracle.com/javase/8/docs/api/java/net/DatagramPacket.html) this
> public void setLength(int length)
Set the length for this packet. The length of the packet is the number of bytes from the packet's data buffer that will be sent, or the number of bytes of the packet's data buffer that will be used for receiving data. The length must be lesser or equal to the offset plus the length of the packet's buffer.

But [it is occurred](https://github.com/jitsi/ice4j/pull/168/files#diff-101ada981bdef99fff62750208d24741R89) that documentation is slightly misleading. There is only one way (I was found it from test) to limit the size of received datagram - is to specify the buffer length [in constructor](https://docs.oracle.com/javase/8/docs/api/java/net/DatagramPacket.html#DatagramPacket-byte:A-int-int-) or via [`void setData(byte[] buf, int offset, int length)`](https://docs.oracle.com/javase/8/docs/api/java/net/DatagramPacket.html#setData-byte:A-int-int-). Changing length via `setLength` before call to `receive` will not limit the size of received data, but if `DatagramPacket` is used for sending it will affect the number of bytes which are sent.
I've added [a unit test](https://github.com/jitsi/ice4j/blob/d160de03848f4d4a2c0617a8d7f53a1f2ff8894d/src/test/java/org/ice4j/util/DatagramPacketBufferTest.java) to verify expected behavior, so further code readers can easily play with sockets and check behavior themselves.

> How about reseting both length and offset ?

There is no way to reset `offset` (no `setOffset` method). 
As I understand `offset` can not be changed neither by `receive` nor by `send`. I believe the reason why datagram packet have such strange signature with `buffer`, `offset` and `length` is to allow programmer to create several `DatagramPacket` instances, which all mapped to single underlying byte array, but with non-overlapping regions specified by `offset` & `length`.


Summarizing, according to [the unit test](https://github.com/jitsi/ice4j/blob/d160de03848f4d4a2c0617a8d7f53a1f2ff8894d/src/test/java/org/ice4j/util/DatagramPacketBufferTest.java) the `setLength` call I've added is not necessary (at least when I run on my machine), but that might be an implementation details. If you want I can remove it.

src/test/java/org/ice4j/util/DatagramPacketBufferTest.java
```diff
@@ -34,7 +34,7 @@ public void testDatagramPacketOffsetAndLengthRespected() throws Exception
         // holds flag if all checks were succeeded.
         final AtomicBoolean allChecksSucceeded = new AtomicBoolean(true);
 
-        final int MAX_SEND_DATAGRAM_SIZE = 256;
+        final int maxSendDatagramSize = 256;
```
paweldomas: I think the preference is for capital letters in constants

src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
paweldomas: I see, thanks

src/test/java/org/ice4j/util/DatagramPacketBufferTest.java
mstyura: Ok, just thought that scope of this constant is a method where lost of other final variables declared. No problem change this back.

src/main/java/org/ice4j/socket/MultiplexingXXXSocketSupport.java
paweldomas: The unit test you've added is nice, but is more like a local experiment as it doesn't test the ice4j itself so I don't think it should land in master.

paweldomas: Can you remove the extra commits related to the unit test and force push ?

PR #170, opened 2019-02-06
Offload I/O from ScheduledExecutorService thread in StunKeepAliveRunner

mstyura:
**TL; DR**: 
Fixed thread starvation in [StunKeepAliveRunner](https://github.com/jitsi/ice4j/blob/e3e78dcda3f791598f6dde256fbaba8ffc2399a3/src/main/java/org/ice4j/ice/Agent.java#L2685) due to operation with blocking I/O executed on `ScheduledExecutorService`'s thread.

**Problem**:
It was reported by Jitsi Team, that changes introduced in PR #151, #152, #153, #155 introduced regression on `TCP`, when there is a thread starvation caused by blocking I/O executed on fixed-size thread pool.

**Solution**:
Split scheduling (delayed execution) from actual execution into different pools: use `ScheduledExecutorService` as a "timer" and cached thread pool for actual task execution. So when some of executed tasks blocked on I/O or something else, it should not prevent scheduling from work and should not affect other tasks.

### DIFF
--- a/src/main/java/org/ice4j/util/PeriodicRunnable.java
+++ b/src/main/java/org/ice4j/util/PeriodicRunnable.java
@@ -95,14 +95,7 @@ public abstract class PeriodicRunnable
         final Duration delay =
             getDelayUntilNextRun();
 
-        synchronized (syncRoot)
-        {
-            if (running)
-            {
-                return;
-            }
-            scheduleNextRun(delay);
-        }
+        scheduleNextRun(delay);
     }
 
     /**
@@ -144,24 +137,36 @@ public abstract class PeriodicRunnable
      */
     private void scheduleNextRun(Duration delay)
     {
-        if (delay.isNegative())
+        synchronized (syncRoot)
         {
-            running = false;
-            return;
-        }
+            final boolean isRecurrentRun = submittedExecute != null;
+            if (isRecurrentRun && !running)
+            {
+                // was cancelled
+                return;
+            }
 
-        running = true;
+            if (delay.isNegative())
+            {
+                running = false;
+                scheduledSubmit = null;
+                submittedExecute = null;
+                return;
+            }
 
-        if (delay.isZero())
-        {
-            submitExecuteRun();
-        }
-        else
-        {
-            scheduledSubmit = timer.schedule(
-                this::submitExecuteRun,
-                delay.toNanos(),
-                TimeUnit.NANOSECONDS);
+            running = true;
+
+            if (delay.isZero())
+            {
+                submitExecuteRun();
+            }
+            else
+            {
+                scheduledSubmit = timer.schedule(
+                    this::submitExecuteRun,
+                    delay.toNanos(),
+                    TimeUnit.NANOSECONDS);
+            }
         }
     }
 
@@ -175,7 +180,14 @@ public abstract class PeriodicRunnable
         {
             return;
         }
-        submittedExecute = this.executor.submit(this::executeRun);
+        synchronized (syncRoot)
+        {
+            if (!running)
+            {
+                return;
+            }
+            submittedExecute = this.executor.submit(this::executeRun);
+        }
     }
 
     /**
@@ -200,13 +212,7 @@ public abstract class PeriodicRunnable
                 final Duration delayMillis =
                     getDelayUntilNextRun();
 
-                synchronized (syncRoot)
-                {
-                    if (running)
-                    {
-                        scheduleNextRun(delayMillis);
-                    }
-                }
+                scheduleNextRun(delayMillis);
             }
         }
     }
--- a/src/main/java/org/ice4j/util/PeriodicRunnable.java
+++ b/src/main/java/org/ice4j/util/PeriodicRunnable.java
@@ -1,5 +1,6 @@
 package org.ice4j.util;
 
+import java.time.*;
 import java.util.concurrent.*;
 
 /**
@@ -73,7 +74,7 @@ public abstract class PeriodicRunnable
      * performed with specified delay, negative value if execution should not
      * be done.
      */
-    protected abstract long getMillisecondsDelayUntilNextRun();
+    protected abstract Duration getDelayUntilNextRun();
 
     /**
      * Periodically executed method on {@link #executor}'s thread.
@@ -91,8 +92,8 @@ public abstract class PeriodicRunnable
             return;
         }
 
-        final long delayMillis =
-            getMillisecondsDelayUntilNextRun();
+        final Duration delay =
+            getDelayUntilNextRun();
 
         synchronized (syncRoot)
         {
@@ -100,7 +101,7 @@ public abstract class PeriodicRunnable
             {
                 return;
             }
-            scheduleNextRun(delayMillis);
+            scheduleNextRun(delay);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/ice4j/util/PeriodicRunnable.java
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * A timer to perform periodic scheduling of {@link #run()} execution
+     * on {@link #executor}'s thread.
+     */
+    private final ScheduledExecutorService timer;
+
+    /**
+     * An executor service to perform actual execution of {@link #run()}.
+     */
+    private final ExecutorService executor;
```
bbaldino: i'm thinking maybe we shouldn't complicate this class with this second executor.  if the caller knows the scheduled task is CPU-bound or IO-bound, it should include submitting the job to the proper scheduler in the submitted task itself.  

In fact, this is making me wonder if this class is really necessary.  why don't we have callers just do:

```
Future<?> scheduledTaskHandle = scheduledExecutorService.scheduleAtFixedRate(..., () -> { cpuPool.submit(doActualWork) }); 
```
or something to that effect?  And when they want to cancel it they just cancel using the handle? do we need all the structure of `PeriodicRunnable`?

src/main/java/org/ice4j/util/PeriodicRunnable.java
```diff
@@ -0,0 +1,248 @@
...
+        this.executor = executor;
+    }
+
+    /**
+     * Get delay before next execution of {@link #run()}.
+     * @return non-negative value if execution of {@link #run()} should be
+     * performed with specified delay, negative value if execution should not
+     * be done.
+     */
+    protected abstract long getMillisecondsDelayUntilNextRun();
```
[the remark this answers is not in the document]
mstyura: Sounds good, will update

src/main/java/org/ice4j/util/PeriodicRunnable.java
```diff
@@ -0,0 +1,248 @@
...
+
+import java.util.concurrent.*;
+
+/**
+ * A base class for runnables which should be periodically executed on
+ * specified executor service.
+ *
+ * @author Yura Yaroshevich
+ */
+public abstract class PeriodicRunnable
```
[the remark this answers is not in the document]
mstyura: Yes, also noted that there is a class with same name, but it has different implementation based on custom single threaded executor service. This was not exact thing what I needed and it isn’t available in ice4j anyway.

src/main/java/org/ice4j/util/PeriodicRunnable.java
mstyura: I’ve also considered this option, but decided to stick with current implementation based on 2 executor services, because it allows more or less robust cancellation - runnable can be cancelled while in timer’s queue, as well as when already on executor’s pool queue. This behavior was before and I don’t want to break it, so i’ve put managing scheduling and cancellation into base class which handles all the details. The same cancellation pattern is used in other places, but with “inlined” implementation. I could eliminate this class or use single executor (it useless with single executor actually as you’ve already noted), but it would complicate cancellation where it used.

bbaldino: True, but I wonder how valuable this is.  There is always a chance a task started running just before we wanted to cancel it, so it's a scenario we have to handle anyway, no?

mstyura: In such case as I understand cancellation relied on interrupt exception caused by `mayInterruptIfRunning - true if the thread executing this task should be interrupted; otherwise, in-progress tasks are allowed to complete`. The value of the class is to don’t repeat yourself, I know at least 3 places where I’ll had to write very similar code to handle non-overlapping in time periodic tasks with cancellation support. Of course I could in-line this class if you insist, just tell me to do so — the thread starvation has to be fixed somehow.

src/main/java/org/ice4j/util/PeriodicRunnable.java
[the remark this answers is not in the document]
bgrozev: +1 for a jitsi-utils package (long term). I hate how libjitsi and jicofo have ice4j as a dependency

src/main/java/org/ice4j/util/PeriodicRunnable.java
```diff
@@ -0,0 +1,242 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * Perform either cancellation or actual scheduling based on current
+     * delay until next run.
+     */
+    private void scheduleNextRun()
```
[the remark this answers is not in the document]
mstyura: @bgrozev sounds good. I've added lock inside method itself. Thank you very much for review!

src/main/java/org/ice4j/util/PeriodicRunnable.java
bbaldino: This logic is needed anyway, is it not? There's always a chance we want to cancel just after a task has started running and we need to be able to handle that.  Or maybe I'm misunderstanding what you're referring to.

mstyura: I feel like we misunderstanding each other :).

Let’s leave cancellation for a moment. Another important property of this class is that currently it simulates behavior of `scheduleWithFixedDelay` (but delay actually can be dynamic) when another pool is used to perform actual job, which I believe is correct for this case: task is not executed again, until previous execution completed - so there is no overlap in time (concurrent) execution of same task. It also allow  change of time interval between each run - which will be useful in a transaction retransmitter (further PR).


Regarding cancellation, there is “best effort” to implement it before task is actually running and yes when task is already started I hard to stop it without proper implementation of task itself. For example, in .NET world the cancellation of running task is done via so called “cooperative cancellation”, when cancellable task accepts so called “CancellationToken”, which is periodically checked by task itself if it cancelled. I don’t know if there is something similar in standard Java library to implement running task cancellation, except `Thread.isInterrupted`... Is it even close to what you are talked about or I still misunderstand you?

bbaldino: > Regarding cancellation, there is “best effort” to implement it before task is actually running and yes when task is already started I hard to stop it without proper implementation of task itself. For example, in .NET world the cancellation of running task is done via so called “cooperative cancellation”, when cancellable task accepts so called “CancellationToken”, which is periodically checked by task itself if it cancelled. I don’t know if there is something similar in standard Java library to implement running task cancellation, except `Thread.isInterrupted`... Is it even close to what you are talked about or I still misunderstand you?

Yeah, I think we understand one another on this point.  My main thought is just: if we have to handle a task running when we want to cancel anyway, why add more code/complexity to try and prevent it from happening if it's impossible to prevent all cases anyway?

> Let’s leave cancellation for a moment. Another important property of this class is that currently it simulates behavior of `scheduleWithFixedDelay` (but delay actually can be dynamic) when another pool is used to perform actual job, which I believe is correct for this case: task is not executed again, until previous execution completed - so there is no overlap in time (concurrent) execution of same task. 

right, ScheduledExecutorService#scheduleWithFixedDelay does exactly this.

> It also allow change of time interval between each run - which will be useful in a transaction retransmitter (further PR).

But why not just accomplish this via a task which is initially scheduled via `ScheduleExecutorService.schedule` and then re-schedules itself via another call to `ScheduleExecutorService.schedule` when it's done?

Btw, if the answer to all this is "because doing it this way results in the fewest changes" I think I'm fine with it for not and we'll plan on cleaning it up later, but since this is an entirely new class I'm wondering if that's the case.

mstyura: Yes, I believe that this class reduces number of changes in this and further PRs with fixes of similar issues.

PR #173, opened 2019-02-12
Offload I/O from ScheduledExecutorService thread in PaceMaker.

mstyura:
**TL; DR**: 
Fixed thread starvation in [PaceMaker](https://github.com/jitsi/ice4j/blob/0be8f1aae79c0bc8efe75f3171345cb8e041d87a/src/main/java/org/ice4j/ice/ConnectivityCheckClient.java#L872) due to operation with blocking I/O executed on `ScheduledExecutorService`'s thread which was introduced #152.

**Problem**:
It was reported by Jitsi Team, that changes introduced in PR #151, #152, #153, #155 introduced regression on `TCP`, when there is a thread starvation caused by blocking I/O executed on fixed-size thread pool.

**Solution**:
Split scheduling (delayed execution) from actual execution into different pools: use `ScheduledExecutorService` as a "timer" and cached thread pool for actual task execution. So when some of executed tasks blocked on I/O or something else, it should not prevent scheduling from work and should not affect other tasks.

### DIFF
--- a/src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
+++ b/src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
@@ -18,9 +18,9 @@
 package org.ice4j.ice;
 
 import java.net.*;
+import java.time.*;
 import java.util.*;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.*;
 import java.util.logging.*;
 
 import org.ice4j.*;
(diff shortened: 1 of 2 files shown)


mstyura:
Jenkins test this please

src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
```diff
@@ -57,10 +58,15 @@
     private final Agent parentAgent;
 
     /**
-     * A scheduled executor service to perform scheduled task of the client
+     * A scheduled executor service to perform periodic tasks of the client
      */
     private final ScheduledExecutorService scheduledExecutorService;
 
+    /**
+     * A scheduled executor service to perform background tasks of the client
```
bbaldino: fix this comment?

mstyura: Oops, fixed.

PR #178, opened 2019-04-16
Update jain-sip version and remove NistSdpFactory dependency

keepsimple1:
Hi, I am a new user of `ice4j` and this is my first pull request.  There are two changes:

1. Upgrade `jain-sip` to more recent version 1.3.0-91 in Maven repo. It helps when the application is using more recent `jain-sip` version.
2. Remove the dependency of `java-sdp-nist-bridge` from `org.opentelecoms.sdp` because it is no longer needed after moving to the newer version of `jain-sip`.  (Note that `SdpFactory` is a Singleton now instead of an abstract class). 

All existing tests passed. Please review and merge. Thank you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@
 .project
 .settings/
 bin/
+
+# Ignore IntelliJ IDEA files
+.idea
+*.iml
--- a/pom.xml
+++ b/pom.xml
@@ -21,14 +21,9 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.opentelecoms.sdp</groupId>
-      <artifactId>java-sdp-nist-bridge</artifactId>
-      <version>1.1</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jitsi</groupId>
-      <artifactId>jain-sip-ri-ossonly</artifactId>
-      <version>1.2.98c7f8c-jitsi-oss1</version>
+      <groupId>javax.sip</groupId>
+      <artifactId>jain-sip-ri</artifactId>
+      <version>1.3.0-91</version>
     </dependency>
     <dependency>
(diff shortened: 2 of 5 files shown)


keepsimple1:
Thanks for your response. I've signed your CLA.

ibauersachs:
Stop, no.
jain-sip-ri contains non-free files and cannot be used (i.e. the SIP API). opentelecoms has reimplemented this under a free license, and the bridge is used to instantiate classes of the free reference implementation.

If at all, the -ossonly package should be updated and uploaded to Maven.

keepsimple1:
Oh, I was not aware of that. Thanks for letting me know.  

There is a slightly newer version of -ossonly package (1.2.279) on Maven: https://mvnrepository.com/artifact/org.jitsi/jain-sip-ri-ossonly   .  Should we use that one?

ibauersachs:
Unless there's something specific that was fixed on those few commits, there's not much point. We should update jitsi/jain-sip the current upstream and do a new release. However, the repo is forked from RestComm/jain-sip, but now NIST has moved their SVN to Github as well and there are now two competing Git repositories. The jain-sip-ri on Maven AFAIK is from the RestComm-Repo.

keepsimple1:
I have a question: isn't jitsi/jain-sip repo still having "javax" files that are under different license (unlike the "gov.nist" files) ?  https://github.com/jitsi/jain-sip/tree/master/src/javax .  How different is it from RestComm/jain-sip so that it makes jitsi/jain-sip fully open source and free?

ibauersachs:
The repo still has these files but they aren't used to produce the final artifact.

keepsimple1:
Got it, just noticed that too. Thanks. I will close this pull request and work with -oss-only for now.

PR #179, opened 2019-04-17
Fix Retransmitter creation to take effect of originalWaitInterval config

keepsimple1:
Hi,

I found that `originalWaitInterval` configuration does not really take effect for `Retransmitter` in `StunClientTransaction`.  The reason was that `retransmitter` was initialized at its declaration instead of in the constructor of `StunClientTransaction`. 

Changes:
1. I've made a change to fix that and added a test following existing testing pattern. 
2. Also it seems that the test class `DatagramCollector` implicitly depends on the jave.net.DatagramSocket locking the packet. I've added a comment for that.  (And in theory there is a issue with Thread.start() timing as well in this test class `DatagramCollector`, but it seems ok with existing tests. ) 

Please review, thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@
 .project
 .settings/
 bin/
+
+# Ignore IntelliJ IDEA files
+.idea
+*.iml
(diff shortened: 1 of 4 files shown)


keepsimple1:
Just to confirm, I've already signed CLA.

keepsimple1:
Could anyone please review this pull request? Thanks.

bgrozev:
Nice catch! Thanks for the fix

PR #180, opened 2019-04-26
Prepares for java 11.

damencho:


src/main/java/org/ice4j/socket/jdk8/DelegatingServerSocketChannel.java
```diff
@@ -54,6 +75,37 @@ public DelegatingServerSocketChannel(T delegate)
 
         delegateAsSelChImpl
             = (delegate instanceof SelChImpl) ? (SelChImpl) delegate : null;
+
+        Method method;
+        try
+        {
+            method = SelChImpl.class.getMethod(
```
bgrozev: Do we want to the the method from `SelChImpl`, or from the class of our delegate? I.e. if the delegate overrides the method, which one should we call?

src/main/java/org/ice4j/socket/jdk8/DelegatingServerSocketChannel.java
```diff
@@ -18,10 +18,13 @@
 package org.ice4j.socket.jdk8;
 
 import java.io.*;
+import java.lang.reflect.*;
 import java.nio.channels.*;
+import java.util.logging.*;
 
 import org.ice4j.socket.*;
 
+import org.ice4j.util.Logger;
```
bgrozev: Is this used anywhere? The `classLogger` is a `java.util.logging.Logger` instance

src/main/java/org/ice4j/socket/jdk8/DelegatingServerSocketChannel.java
damencho: I need to test it, as there was some problem which I do not recall well when getting the class from the instance ...

damencho: It doesn't work obtaining the method from the instance with delegateAsSelChImpl.getClass().getMethod. The error:
java.lang.IllegalAccessException: class org.ice4j.socket.jdk8.DelegatingServerSocketChannel cannot access a member of class sun.nio.ch.ServerSocketChannelImpl (in module java.base) with modifiers "public"

With obtaining the method from the class it runs with one warning in the beginning:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.ice4j.socket.jdk8.DelegatingServerSocketChannel (file:/usr/share/jitsi-videobridge/lib/ice4j-2.0.0-SNAPSHOT.jar) to method sun.nio.ch.SelChImpl.translateInterestOps(int)
WARNING: Please consider reporting this to the maintainers of org.ice4j.socket.jdk8.DelegatingServerSocketChannel
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

```

src/main/java/org/ice4j/socket/jdk8/DelegatingSocketChannel.java
```diff
@@ -53,6 +73,37 @@ public DelegatingSocketChannel(T delegate)
...
+            method = SelChImpl.class.getMethod(
+                "translateAndSetInterestOps",
+                Integer.TYPE, SelectionKeyImpl.class);
+        }
+        catch(NoSuchMethodException e)
+        {
+            method = null;
+
+            classLogger.log(
+                Level.SEVERE,
```
bkempe: This will log a `SEVERE` level message in either Java 8 or Java 11. Maybe it would be better to only log if neither method is found?

damencho: Yeah, you are right I saw it today ... maybe we will revert it cause it failed to work when only java11 is installed on the system, not sure why. 
Testing before that with java8 and 11 installed and 11 to be default this was working.

bkempe: Is Java11 support still work-in-progress?

damencho: Yes, any help is welcome.

src/main/java/org/ice4j/socket/jdk8/DelegatingServerSocketChannel.java
deepdad: I get this in the startup log (Illegal reflective access operation) with openjdk 11.0.6 on Ubuntu18.04

PR #183, opened 2019-06-07
Removes extra exception printed on initialising delegating channels.

damencho:


src/main/java/org/ice4j/socket/jdk8/DelegatingServerSocketChannel.java
```diff
@@ -101,10 +97,16 @@ public DelegatingServerSocketChannel(T delegate)
...
-                Level.SEVERE, "Cannot find method translateInterestOps", e);
         }
         translateInterestOpsMethod = method;
+
+        if (translateInterestOpsMethod == null
+            && translateAndSetInterestOpsMethod == null)
+        {
+            classLogger.log(
+                Level.SEVERE, "Cannot find translateInterestOps " +
+                    "or translateAndSetInterestOps!");
```
bgrozev: The log says "or", but the condition of the if uses "and". I think it should be "or" in both places

damencho: Nope, it is correct. If we return the or in the if, we will again see the exception every time, cause every time there will be one of the methods not found.

The thing is that when you run with java8 you will have method A and when you run with java 11 you will have method B. And we want to print an error only in case there is none found.

bgrozev: Ah, I see

PR #184, opened 2019-09-14
increase major version number

bbaldino:
@bgrozev and i have discussed making some changes in ice4j that may or may not work for jitsi desktop (or, at least, not in the same timeframe we want to use them elsewhere).  after talking with @damencho as well, we decided a plan forward would be to cut ice4j's current master to a branch (which can be found [here](https://github.com/jitsi/ice4j/tree/v2.x)) and changing master over to 3.0.  We can continue to deploy the 2.0 branch to sonatype and Jitsi desktop can use the 2.0 versions. The 3.0 (master branch) will be deployed to the jitsi repo.  

Appropriate changes can be ported from master to the 2.0 branch, but it would not necessarily get every change from 3.0 (the new master).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
   <groupId>org.jitsi</groupId>
   <artifactId>ice4j</artifactId>
-  <version>2.0.0-SNAPSHOT</version>
+  <version>3.0-SNAPSHOT</version>
   <packaging>bundle</packaging>
 
   <name>ice4j</name>


bbaldino:
> I'd prefer if you stick to semver - i.e. use 3.0.0-SNAPSHOT. But since we never actually released 2.0.x, it doesn't really matter.
> 

This scheme (`3.0-SNAPSHOT`) is what @damencho did in the other components--I think because we don't explicitly set a patch number, we pull it from git--so I was consistent with that here.


> What are those changes you're talking about that ice4j can't be updated in Jitsi/Jigasi anymore?

I don't know if there's anything now that we _know_ couldn't be merged, but we'd like to be able to keep things moving without having to worry about breaking some of the existing components (especially Jitsi desktop since we don't really do any testing there) to use new features in other things that use ice4j.

ibauersachs:
ice4j was the only lib that I tried to do semver as there are users outside of the Jitsi ecosystem. But as I said, there was never actually a 2.x release, so people probably arranged around the pinned snapshots or now the commit-hashed versions, so it really doesn't matter. Especially if nobody commits to keeping the lib on Maven Central up to date and keeping an eye on increasing the versions by feature/patch.

bbaldino:
My understanding is that no one here knows how to push this to sonatype/maven central (or maybe there are some credentials that we don't have?) and that you were the one doing that.  Is that right?

ibauersachs:
I pushed 1.0, yes. There are a few people with permissions to push to Sonatype (I and Damencho for sure, probably Boris, Emcho and Lyubo too), and how-to is just as any other Maven project, i.e. mvn publish.
The jenkins user has permissions to push -snapshots only (but that could be lifted if necessary or another service account added with full permissions).

bbaldino:
Ok, sounds like there may be a disconnect but I don't have all the info.  If we can get that sorted then pushing releases from the from the `ice4j_2.0` branch automatically probably wouldn't be too bad.

bbaldino:
should i change the pom in the `ice4j_2.0` branch to be `2.0-SNAPSHOT` then as well? Or leave it?

ibauersachs:
Leave it at 2.0.0-snapshot, but ideally rename the branch to v2.x (just like v1.x is already there).

bbaldino:
> Leave it at 2.0.0-snapshot, but ideally rename the branch to v2.x (just like v1.x is already there).

Done.  You all good with merging this then?

PR #186, opened 2019-09-17
use the jitsi-utils logger

bbaldino:


ibauersachs:
Why can't we use slf4j, especially when touching a lot of logging anyway?

bbaldino:
> Why can't we use slf4j, especially when touching a lot of logging anyway?

There are 2 main questions here: one is the the log context and the other is the enforcement of the minimum log level from the top down.  The log context stuff slf4j does have support for and making it work with java.util.logger it looks like it should be doable? But I haven't taken a look at it.  The other feature is kinda unique and I'm not sure if slf4j supports it (basically we need to enforce that a parent logger and all of its children don't log messages below a certain level, even if they are configured to).  

If you've got any insight on those I'd be interested to hear it.  I'll admit, though, that the main reason I didn't change to slf4j or dig too deeply into whether or not it supports these things is that, as big as these changes may seem, they're still entirely superficial and I didn't want to (lazy?) get into the work of validating a change of logging framework.

bbaldino:
sorry, have to close this because the branch name didn't match that used in the ice4j PR so the tests don't pass.  https://github.com/jitsi/jitsi-videobridge/pull/918 is the new one.

bbaldino:
jenkins, test this please

bbaldino:
so i think it may be impossible to get the tests passing here.  jicofo fails to build because it relies on the bridge, but we can't update the bridge dependency because we don't have one until we have an ice4j version...which we won't get until this PR merges :)

src/main/java/org/ice4j/ice/LocalCandidate.java
```diff
@@ -91,12 +90,10 @@ public LocalCandidate(TransportAddress transportAddress,
                           CandidateType    type,
                           CandidateExtendedType extendedType,
                           LocalCandidate  relatedCandidate)
+
```
virtuacoplenny: nit: extra line break?

damencho:
Hum, all with same branch should use those branches as dependencies, shouldn't that work?

damencho:
The problem is that the PR testing scripts does not take care of jitsi-utils at all.

damencho:
Seems to be fixed now, but for some reason in jicofo logs I see it uses `/home/jenkins/.m2/repository/org/jitsi/ice4j/2.0.0-SNAPSHOT/ice4j-2.0.0-SNAPSHOT.jar` still not sure why.

damencho:
Created 2 dummy PRs in jicoco and jitsi in order to update jicofo dependency to ice4j when testing using the videobridge...

PR #187, opened 2019-09-20
Added `Message-Integrity` attribute to `Role Conflict` message.

mstyura:
When `ICE` role conflict happens, `ice4j` sends `Role Conflict` message, which is not liked by both `Chrome` and `Firefox`.
Firefox seems to drops such `STUN` error response, producing following logs:
```
(ice/INFO) ICE-PEER(PC:1568974392290000 (id=40802189545 url=https://127.0.0.1:9980/):default)/CAND-PAIR(pX/O): setting pair to state IN_PROGRESS: pX/O|IP4:10.0.75.1:60420/UDP|IP4:10.0.75.1:10000/UDP(host(IP4:10.0.75.1:60420/UDP)|candidate:1 1 udp 2130706431 10.0.75.1 10000 typ host generation 0)

(stun/INFO) STUN-CLIENT(pX/O|IP4:10.0.75.1:60420/UDP|IP4:10.0.75.1:10000/UDP(host(IP4:10.0.75.1:60420/UDP)|candidate:1 1 udp 2130706431 10.0.75.1 10000 typ host generation 0)): Received response; processing

(stun/WARNING) Missing MESSAGE-INTEGRITY

(stun/WARNING) STUN-CLIENT(pX/O|IP4:10.0.75.1:60420/UDP|IP4:10.0.75.1:10000/UDP(host(IP4:10.0.75.1:60420/UDP)|candidate:1 1 udp 2130706431 10.0.75.1 10000 typ host generation 0)): Error processing response: Operation rejected, stun error code 0.
```

`Chrome` seems to do the same, but [silenty](https://cs.chromium.org/chromium/src/third_party/webrtc/p2p/base/connection.cc?rcl=fb59a6aa3f718241506fb644f4daa295f992d596&l=457).

According to `Chrome` [source code](https://cs.chromium.org/chromium/src/third_party/webrtc/p2p/base/port.cc?rcl=fb59a6aa3f718241506fb644f4daa295f992d596&l=749) `Message-Integrity` is mandatory attribute for `Role Conflict` message.

With this fix, `Chrome` properly handles `Role Conflict` response and switched it's role.
`Firefox` for some reason still unable to handle `STUN` error response to recover from `Role Conflict`, but have different message in logs:
```
(ice/INFO) ICE-PEER(PC:1568984401039000 (id=40802189569 url=https://127.0.0.1:9980/):default)/CAND-PAIR(8hTQ): setting pair to state IN_PROGRESS: 8hTQ|IP4:10.0.75.1:52398/UDP|IP4:10.0.75.1:10000/UDP(host(IP4:10.0.75.1:52398/UDP)|candidate:1 1 udp 2130706431 10.0.75.1 10000 typ host generation 0)

(stun/INFO) STUN-CLIENT(8hTQ|IP4:10.0.75.1:52398/UDP|IP4:10.0.75.1:10000/UDP(host(IP4:10.0.75.1:52398/UDP)|candidate:1 1 udp 2130706431 10.0.75.1 10000 typ host generation 0)): Received response; processing

(stun/WARNING) STUN-CLIENT(8hTQ|IP4:10.0.75.1:52398/UDP|IP4:10.0.75.1:10000/UDP(host(IP4:10.0.75.1:52398/UDP)|candidate:1 1 udp 2130706431 10.0.75.1 10000 typ host generation 0)): Error processing response: Retry may be possible, stun error code 487.

(ice/INFO) ICE-PEER(PC:1568984401039000 (id=40802189569 url=https://127.0.0.1:9980/):default)/CAND-PAIR(zpTK): setting pair to state IN_PROGRESS: zpTK|IP4:10.0.75.1:52398/UDP|IP4:172.19.103.121:10000/UDP(host(IP4:10.0.75.1:52398/UDP)|candidate:2 1 udp 2130706431 172.19.103.121 10000 typ host generation 0)
```

I've inspected `Firefox` source code responsible for `ICE` implementation and was failed to find proper role conflict implementation... There are code bits which handles role conflict, but for some reason they were not executed according to logs :(

### DIFF
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -242,7 +242,7 @@ public class Agent
     private long tieBreaker;
 
     /**
-     * Determines whether this is the controlling agent in a an ICE interaction.
+     * Determines whether this agent has a controlling role in an ICE interaction.
      */
     private boolean isControlling = true;
 
@@ -402,7 +402,8 @@ public class Agent
             addCandidateHarvester(harvester);
         }
 
-        logger.debug(() -> "Created a new Agent");
+        logger.debug(() -> "Created a new Agent: " + this.toString() +
+            " with ICE controlling role = " + this.isControlling);
     }
 
     /**
@@ -1407,8 +1408,13 @@ public class Agent
      */
     public void setControlling(boolean isControlling)
     {
-        this.logger.info(() -> "Changing agent " + this.toString() + " role from controlling = "
-            + this.isControlling + " to controlling = " + isControlling);
+        if (this.isControlling != isControlling)
+        {
+            this.logger.info(() -> "Changing agent " + this.toString() +
+                " role from controlling = " + this.isControlling +
+                " to controlling = " + isControlling);
+        }
+
         this.isControlling = isControlling;
 
         //in case we have already initialized our check lists we'd need to
--- a/src/main/java/org/ice4j/ice/Agent.java
+++ b/src/main/java/org/ice4j/ice/Agent.java
@@ -1407,6 +1407,8 @@ public class Agent
      */
     public void setControlling(boolean isControlling)
     {
+        this.logger.info(() -> "Changing agent " + this.toString() + " role from controlling = "
+            + this.isControlling + " to controlling = " + isControlling);
         this.isControlling = isControlling;
 
         //in case we have already initialized our check lists we'd need to
--- a/src/main/java/org/ice4j/ice/ConnectivityCheckServer.java
+++ b/src/main/java/org/ice4j/ice/ConnectivityCheckServer.java
@@ -302,7 +302,7 @@ class ConnectivityCheckServer
             }
         }
         //If the agent's tie-breaker is less than the contents of the
-        //ICE control attribute, the agent toggles it's ICE control role.
+        //ICE control attribute, the agent toggles its ICE control role.
         else
         {
             final String selfNextControlState
--- a/src/main/java/org/ice4j/ice/ConnectivityCheckServer.java
+++ b/src/main/java/org/ice4j/ice/ConnectivityCheckServer.java
@@ -36,23 +36,6 @@ class ConnectivityCheckServer
                CredentialsAuthority
 {
     /**
-     * Compares <tt>a</tt> and <tt>b</tt> as unsigned long values. Serves the
-     * same purpose as the <tt>Long.compareUnsigned</tt> method available in
-     * Java 1.8.
-     * @return <tt>-1</tt> if <tt>a</tt> is less than <tt>b</tt>, <tt>0</tt> if
-     * they are equal and <tt>1</tt> if <tt>a</tt> is bigger.
-     */
-    private static int compareUnsignedLong(long a, long b)
-    {
-        if (a == b)
-            return 0;
-        else if ((a + Long.MIN_VALUE) < (b + Long.MIN_VALUE))
-            return -1;
-        else
-            return 1;
-    }
-
-    /**
      * The agent that created us.
      */
     private final Agent parentAgent;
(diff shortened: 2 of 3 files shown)


mstyura:
Hello @bbaldino!

Could you please review these change? 

Thanks a lot!

src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -1407,6 +1407,8 @@ public long getTieBreaker()
      */
     public void setControlling(boolean isControlling)
     {
+        this.logger.info(() -> "Changing agent " + this.toString() + " role from controlling = "
+            + this.isControlling + " to controlling = " + isControlling);
```
JonathanLennox: It seems like this will log not only for role changes (which is good), but also on ordinary startup for a controlled node (which will be a lot of unnecessary noise).

src/main/java/org/ice4j/ice/ConnectivityCheckServer.java
```diff
@@ -259,107 +242,78 @@ private long extractPriority(Request request)
...
-                    //rethrow so that we would send a 500 response instead.
-                    throw new RuntimeException("Failed to send a 487", exc);
-                }
+                //rethrow so that we would send a 500 response instead.
+                throw new RuntimeException("Failed to send a 487", exc);
             }
         }
-        return true; // we don't have a role conflict
+        //If the agent's tie-breaker is less than the contents of the
+        //ICE control attribute, the agent toggles it's ICE control role.
```
JonathanLennox: Should be "its" role.

bbaldino:
@mstyura would it be possible to add a unit test to validate the conflict case?

src/main/java/org/ice4j/ice/Agent.java
mstyura: @JonathanLennox do you mean you want `if (this.isControlling != isControlling)` before printing log?

JonathanLennox: No, I think what I mean is that ideally, the first time this is set, the "from" part of the log message wouldn't be included (since I feel it's not particularly meaningful in that case).

This would add state complexity which would only be used in log messages, though, so I'd be willing to let @bbaldino disagree with me.

mstyura: @JonathanLennox ok, I think I've got generic idea.
What do you think of updating this line from constructor:
https://github.com/jitsi/ice4j/blob/e854e4efaf5dbb82e45028fceafe9d76ca93cb2a/src/main/java/org/ice4j/ice/Agent.java#L405
To something like this:
```
logger.debug(() -> "Created a new Agent: " + this.toString() + " with ICE controlling role = " + this.isControling);
```
And in `setControlling` I could also add `if (this.isControlling != isControlling)` before printing. So it should solve the issue of unnecessary print to log when nothing is changed.

BTW, I could also update logging level to debug, so logs will only be polluted when detailed logs are enabled. What do you think?

JonathanLennox: The problem with both of these is that `isControlling` isn't set in the constructor, it just defaults to `true`.  So for a controlled endpoint you'd always get "Created a new Agent: ... with ICE controlling role = true" and then "Changing Agent from controlling = true to controlling = false", every time.  This is what seems misleading to me.

mstyura: @JonathanLennox then the only way to resolve this issue seems to make initial ICE role as required constructor argument :) I could do this, but I will need to update JVB side to leverage new constructor (I'm actually working on PR for JVB which adjusts `Dtls/IceTransport` implementation, so it shouldn't be a problem at all to use potentially new `Agent`'s constructor) . If you think that it would be a good solution, I'll implement it, just let me know please. Thank you very much for review!

bbaldino: i think the idea is sound, but given there doesn't seem to be a clean solution to get a better behavior, i think i'd rather leave it as-is than get into changing the ctor argument.  (i do, however, think that doing `if (this.isControlling != isControlling)` before printing log makes sense)

bbaldino:
thanks @mstyura , it's unfortunate the test has to be so broad but i do think it's useful, nice job on that and thanks for adding it (even with the caveat).

mstyura:
@bbaldino can you merge this PR if it's ok?

mstyura:
Hi, @bbaldino,
Do you think there there is anything else need to be fixed in this PR before it can be approved?
Thanks a lot!

bbaldino:
deferred to @JonathanLennox on this one.   lgth :)

PR #191, opened 2020-03-23
Updated java-sdp-nist-bridge and removed explict jain-sip-ri-ossonly dependency.

mstyura:
It seems that removed dependency is unused. 
```
 mvn clean package test
```
completes without errors with that dependency removed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,11 +26,6 @@
       <version>1.1</version>
     </dependency>
     <dependency>
-      <groupId>org.jitsi</groupId>
-      <artifactId>jain-sip-ri-ossonly</artifactId>
-      <version>1.2.98c7f8c-jitsi-oss1</version>
-    </dependency>
-    <dependency>
       <groupId>org.bitlet</groupId>
       <artifactId>weupnp</artifactId>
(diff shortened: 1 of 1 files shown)


ibauersachs:
STOP!

ibauersachs:
I'm not sure anymore how this is exactly used, but probably via reflection in java-sdp-nist-bridge. It is used for the javax.sdp.* stuff, which is still referenced. It's possible that java-sdp-nist-bridge pulls in the dependency anway, or possibly an older version that had bugs with "o=". This needs to be checked more thoroughly than just "it compiles".

mstyura:
@ibauersachs seems like the dependency I've removed is anyway come with `java-sdp-nist-bridge` artifact: 
https://github.com/opentelecoms-org/java-sdp-nist-bridge/blob/c3b3dd6dc5bc8112357d435f3eb05c217b823101/pom.xml#L45-L49
but has slightly different version.
If change does something wrong I'd like to update it/create new PR with comment explanation of such an implicit usage.

mstyura:
Jenkins test this please

PR #195, opened 2020-04-06
Merge duplicate ICE candidate pairs.

JonathanLennox:
This can happen if a remote candidate is discovered as peer-reflexive and later received over signaling.

### DIFF
--- a/src/main/java/org/ice4j/ice/CheckList.java
+++ b/src/main/java/org/ice4j/ice/CheckList.java
@@ -210,6 +210,25 @@ public class CheckList
     }
 
     /**
+     * Returns a pair which matches the specified local and remote candidates'
+     * transport addresses, if any, otherwise null.
+     */
+    public synchronized CandidatePair findPairMatching(LocalCandidate local, RemoteCandidate remote)
+    {
+        for (CandidatePair pair : this)
+        {
+            if (pair.getLocalCandidate() == local ||
+                pair.getLocalCandidate().getTransportAddress().equals(local.getTransportAddress()))
+            {
+                if (pair.getRemoteCandidate() == remote ||
+                    pair.getRemoteCandidate().getTransportAddress().equals(remote.getTransportAddress()))
+                    return pair;
+            }
+        }
+        return null;
+    }
+
+    /**
      * Determines whether this <tt>CheckList</tt> can be considered active.
      * RFC 5245 says: A check list with at least one pair that is Waiting is
      * called an active check list.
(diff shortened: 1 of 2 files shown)


src/main/java/org/ice4j/ice/CheckList.java
```diff
@@ -209,6 +209,25 @@ protected synchronized CandidatePair getNextOrdinaryPairToCheck()
         return highestPriorityPair; //return even if null
     }
 
+    /**
+     * Returns a pair which matches the specified local and remote candidates'
+     * transport addresses, if any, otherwise null.
+     */
+    public synchronized CandidatePair findPairMatching(LocalCandidate local, RemoteCandidate remote)
```
bbaldino, line 216: Why synchronized?  And what about just implementing equals for `CandidatePair`?

JonathanLennox: `CandidatePair` already has `equals`, which tests full equality on the candidates (not just their `TransportAddress`es.

bbaldino:
lgtm, but I'll wait for @bgrozev to weigh in.

src/main/java/org/ice4j/ice/Component.java
```diff
@@ -465,7 +463,24 @@ public void updateRemoteCandidates()
...
+                     * (This can happen for pairs with remote peer-reflexive candidates,
+                     *  since those candidates aren't added to the candidate list even
+                     *  though the pair is added to the check list).
+                     */
+                    CandidatePair existingPair = streamCheckList.findPairMatching(pair.getLocalCandidate(), pair.getRemoteCandidate());
+                    if (existingPair != null) {
+                        logger.info("existing Pair updated: " +
+                            existingPair.toShortString() +
+                            " to " + pair.toShortString() + ".");
+                        existingPair.setRemoteCandidate(pair.getRemoteCandidate());
```
bgrozev, line 476: Should we be smarter about which remote candidate we maintain? I.e. make sure we use whichever one is not prflx? Conceivably if candidates trickle and are delayed in signaling we may get a prflx prior to a signaled one.

JonathanLennox: Ones that come in signaling should always win over ones that were discovered as prflx (this can be determined through deep textual analysis of RFC 5245).

The only way we hit this code is if the check being replaced is on the checklist but its remote candidate isn't on the remote candidate list, i.e. if this check was indeed discovered as prflx.

PR #197, opened 2020-04-07
Some helpers for adding an ICE nomination strategy config option.

JonathanLennox:


src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -1850,6 +1850,18 @@ public synchronized void nominate(CandidatePair pair)
...
     }
 
+    /**
+     * Returns the {@link NominationStrategy} that we use in order to
+     * decide if and when we should nominate valid pairs.
+     *
+     * @return the strategy that we are using for nominating
+     * valid {@link CandidatePair}s.
+     */
+    public NominationStrategy getNominationStrategy()
```
bbaldino, line 1860: I don't see this used in JVB?

JonathanLennox: It's not - but it's what convinces Kotlin that `nominationStrategy` can be treated as a`var`, so we can do the more Kotlin-idiomatic `nominationStrategy = Config.nominationStrategy()` rather than `setNominationStrategy(Config.nominationStrategy())`.

bbaldino: In that case, I'm not sure it's doing what we think it's doing.  Have you verified updates are reflected?  I'd think it'd be 
 compiling down to something like:
```
NominationStrategy ns = Agent.getNominationStrategy();
ns = Config.nominationStrategy();
```
which according to the little test I just tried isn't going to reflect back onto the class' value.

JonathanLennox: Testing it shows the config value taking effect.  See https://kotlinlang.org/docs/reference/java-interop.html#getters-and-setters .

bbaldino: Oh, I didn't realize there was already a setter before this PR.

JonathanLennox: Oh, yeah, there was; that was always there.

PR #212, opened 2020-08-20
Use metaconfig, port some properties.

bgrozev:


src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -2690,21 +2646,14 @@ protected Logger getLogger()
...
     private final class StunKeepAliveRunner extends PeriodicRunnable
     {
-        private final long consentFreshnessInterval = Long.getLong(
-            StackProperties.CONSENT_FRESHNESS_INTERVAL,
-            DEFAULT_CONSENT_FRESHNESS_INTERVAL);
-
-        private final int originalConsentFreshnessWaitInterval = Integer.getInteger(
-            StackProperties.CONSENT_FRESHNESS_ORIGINAL_WAIT_INTERVAL,
-            DEFAULT_CONSENT_FRESHNESS_ORIGINAL_WAIT_INTERVAL);
+        private final long consentFreshnessInterval = config.getConsentFreshnessInterval().toMillis();
```
bbaldino, line 2649: Given that the code still seems to require `long` everywhere, should we have the props be `Long`? (new config can still use a `Duration` but convert to Long rather than converting the legacy config Long to a Duration)

PR #214, opened 2020-09-12
Port more config properties.

bgrozev:


bbaldino (approved):

src/main/kotlin/org/ice4j/ice/harvest/HarvestConfig.kt
```diff
@@ -0,0 +1,63 @@
...
+            .convertFrom<Long> { Duration.ofSeconds(it) }
+        "ice4j.harvest.timeout".from(configSource)
+    }
+    fun timeout() = timeout
+
+    companion object {
+        @JvmField
+        val config = HarvestConfig()
+    }
+}
```
JonathanLennox, line 63: No newline at end of file - indicating we don't have ktlint set up here.

bgrozev: I'll add that to my list and do it in another PR.

PR #218, opened 2020-10-14
update jicoco dep

bbaldino:
see https://github.com/jitsi/jicoco/pull/118


bbaldino:
Don't think this is needed

PR #221, opened 2020-12-20
Migrate tests to JUnit 5 and release Maven Central

ibauersachs:


ibauersachs:
@bgrozev / @bbaldino Sorry, I know this is big, but the parameter change of the assertions unfortunately causes a big churn.

JonathanLennox:
Hm, I realized an issue with this (when I had to re-run the tests of my subsequent PR): we run the ice4j unit tests simultaneously with Java 8 and Java 11. Unfortunately, ice4j has unit tests (in PseudoTcpStreamTest) that bind on a hard-coded port.  If the Java 8 and Java 11 tests happen to hit that test simultaneously, one of them will fail with a bind error.

PR #223, opened 2021-04-08
Shut down harvester thread pool when it's done.

JonathanLennox:


gpolitis (approved):

src/main/java/org/ice4j/ice/harvest/MappingCandidateHarvesters.java
```diff
@@ -319,6 +319,8 @@ public StunMappingCandidateHarvester call()
             }
         }
 
+        es.shutdown();
```
bgrozev, line 322: Can we also name the threads? I've been wondering what "pool-1-thread-1" is for a long time.

ibauersachs:
Why is there a thread pool at all when it is just used once in a synchronous method?

Plus, the shutdown is not invoked when the method fails anywhere (e.g. due to interruption or when any of the harvesters fails).

JonathanLennox:
> Why is there a thread pool at all when it is just used once in a synchronous method?

I didn't write this code, but I assume the goal is to run all the harvesters in parallel.  I guess raw `Thread` calls could be used rather than `ExecutorService`, but it doesn't seem that important.

> Plus, the shutdown is not invoked when the method fails anywhere (e.g. due to interruption or when any of the harvesters fails).

Ah, good point, I'll change that.

ibauersachs:
> > Why is there a thread pool at all when it is just used once in a synchronous method?
> 
> I didn't write this code, but I assume the goal is to run all the harvesters in parallel. I guess raw `Thread` calls could be used rather than `ExecutorService`, but it doesn't seem that important.

How often does this code run and how big is the `tasks` list usually? Creating threads is expensive and might be better to have a static thread pool with a limited amount of threads (say 5) than to create them over and over again.

I didn't write this either, but the amount of threads various projects are throwing around is astonishing. Not for this PR, but going over stuff like this (and making it properly async) would probably be a tremendous performance improvement.

JonathanLennox:
> How often does this code run and how big is the `tasks` list usually? Creating threads is expensive and might be better to have a static thread pool with a limited amount of threads (say 5) than to create them over and over again.

It's only called from `MappingCandidateHarvesters#initialize()`, so it's only called once, the first time `MappingCandidateHarvester#getHarvesters()` is called.  I assume the parallelization is to make startup fast.

The number of tasks is equal to (number of local addresses) * (number of configured stun servers).

src/main/java/org/ice4j/ice/harvest/MappingCandidateHarvesters.java
```diff
@@ -283,42 +283,53 @@ public StunMappingCandidateHarvester call()
 
         // Now run discover() on all created harvesters in parallel and pick
         // the ones which succeeded.
-        ExecutorService es = Executors.newFixedThreadPool(tasks.size());
+        ExecutorService es = Executors.newFixedThreadPool(tasks.size(),
+            runnable -> {
+                Thread thread = new Thread(runnable);
+                thread.setName("ICE Harvester Executor");
```
ibauersachs, line 289: You'll have `tasks.size()` threads named `ICE Harvester Executor` with this. The default factory takes care of that: `pool-1-thread-1`. Again, depends on the size of the tasks list if this is relevant.

src/main/java/org/ice4j/ice/harvest/MappingCandidateHarvesters.java
```diff
@@ -283,42 +283,53 @@ public StunMappingCandidateHarvester call()
 
         // Now run discover() on all created harvesters in parallel and pick
         // the ones which succeeded.
-        ExecutorService es = Executors.newFixedThreadPool(tasks.size());
+        ExecutorService es = Executors.newFixedThreadPool(tasks.size(),
+            runnable -> {
+                Thread thread = new Thread(runnable);
```
ibauersachs, line 288: You may want to set this thread as a daemon thread. See the explanation here: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html#defaultThreadFactory()

ibauersachs:
> > How often does this code run and how big is the `tasks` list usually? Creating threads is expensive and might be better to have a static thread pool with a limited amount of threads (say 5) than to create them over and over again.
> 
> It's only called from `MappingCandidateHarvesters#initialize()`, so it's only called once, the first time `MappingCandidateHarvester#getHarvesters()` is called. I assume the parallelization is to make startup fast.

Okay, since this is a one-off thing, don't bother. The other two inline comments might still apply though.

src/main/java/org/ice4j/ice/harvest/MappingCandidateHarvesters.java
```diff
@@ -284,11 +285,18 @@ public StunMappingCandidateHarvester call()
         // Now run discover() on all created harvesters in parallel and pick
         // the ones which succeeded.
         ExecutorService es = Executors.newFixedThreadPool(tasks.size(),
-            runnable -> {
-                Thread thread = new Thread(runnable);
-                thread.setName("ICE Harvester Executor");
-                return thread;
-            });
+            new ThreadFactory() {
```
bgrozev, line 288: You could use [CustomizableThreadFactory](https://github.com/jitsi/jitsi-utils/blob/master/src/main/java/org/jitsi/utils/concurrent/CustomizableThreadFactory.java)

JonathanLennox: Actually even better it looks like we can use `ExecutorService#createFixedThreadPool` which itself uses `CustomizableThreadFactory` -- ice4j is already using the `ExecutorService` methods in a bunch of places.

bgrozev (approved):

PR #225, opened 2021-04-12
feat: Adds a coverage profile and configures Codecov.

gpolitis:


pom.xml
```diff
@@ -287,6 +287,35 @@
   </build>
 
   <profiles>
+    <profile>
+      <id>coverage</id>
+      <properties>
+        <jacoco.version>0.8.6</jacoco.version>
```
ibauersachs, line 293: I know this is copy-paste, but I don't see much value in creating a variable when the value is only used once. Feel free to merge anyway.

gpolitis: Actually that's a good point, creating a variable it's just a waste of bits. I've fixed that now.

damencho (approved):

PR #229, opened 2021-05-20
Port NAT harvester properties to new config

bgrozev:


damencho (approved):

bgrozev:
Thank you for noticing! The intention was to use "local-address" and "public-address" to follow the convention. Fixed in #230. You'll have to update your config when it hits videobridge, but you can add the new names to your config file in advance:
```
static-mappings = [
        {
            localAddress = "1.1.1.1"
            local-address = "1.1.1.1"
            publicAddress = "1.2.3.4"
            public-address = "1.2.3.4"
        }
      ]
```

PR #231, opened 2021-07-12
Enable ephemeral port usage

nils-ohlmeier:
This enables using port=0 in AbstractUdpListener to indicate that a random, ephemeral port should be used instead of a configured port number.

### DIFF
--- a/src/test/java/org/ice4j/ice/harvest/SinglePortUdpHarvesterTest.java
+++ b/src/test/java/org/ice4j/ice/harvest/SinglePortUdpHarvesterTest.java
@@ -115,13 +115,17 @@ public class SinglePortUdpHarvesterTest
     }
 
     @Test
-    public void testBindWithPortZero() throws Exception {
+    public void testBindWithPortZero() throws Exception
+    {
         // Setup test fixture.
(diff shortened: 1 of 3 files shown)


src/test/java/org/ice4j/ice/harvest/SinglePortUdpHarvesterTest.java
```diff
@@ -113,4 +113,25 @@ public void testRebindWithClose() throws Exception
             }
         }
     }
+
+    @Test
+    public void testBindWithPortZero() throws Exception {
```
bgrozev, line 118: nit: In java code we always put braces on a new line (also applies below).

bgrozev:
Jenkins, add to whitelistt

src/test/java/org/ice4j/ice/harvest/SinglePortUdpHarvesterTest.java
nils-ohlmeier: Thanks. I'll fix it in the next push.

Should we add automated linting to ice4j as we have for other projects as well to catch issues like this?

bgrozev (approved):

PR #232, opened 2021-07-14
Update kotlin to 1.5.20.

bgrozev:


ibauersachs:
👏
Time to use only Java 11 as well?

bgrozev:
> Time to use only Java 11 as well?

That would be nice. We're now using 11+ in our deployments. I don't know of any case that requires 8. @damencho ?

ibauersachs:
Well, if you're already using 11 in production, then the best way to enforce that is probably to remove the Java 8 build in jitsi-utils now (but maybe keep the matrix build and add 17-ea instead). The Kotlin upgrade will propagate through everywhere now, so this would be a good chance.

JonathanLennox (approved):

bgrozev:
I don't want to add more to this changeset, but I can do it as a follow up later (after discussing with the team).

ibauersachs:
Yes, that's good. I just wanted to point out the opportunity :-)

JonathanLennox (approved):

PR #233, opened 2021-07-15
feat: Adds checkstyle module.

damencho:


src/main/java/org/ice4j/stack/StunServerTransaction.java
```diff
@@ -165,7 +165,8 @@ public void sendResponse(Response         response,
                IOException,
                IllegalArgumentException
     {
-        if(!isRetransmitting){
+        if(!isRetransmitting)
```
JonathanLennox, line 168: Don't we normally have a space after `if`?  Is checkstyle not catching this?

ibauersachs:
I suggested this before to @damencho not sure if it made it to everyone: https://github.com/google/google-java-format

I know it would be a radical change, but it would squash any discussions about style. Apply the very strict formatter, enforce it with a Maven goal and you're done. The IntelliJ plugin does everything for you with Ctrl+Alt+L.

damencho:
> I suggested this before to @damencho not sure if it made it to everyone: https://github.com/google/google-java-format
> 
> I know it would be a radical change, but it would squash any discussions about style. Apply the very strict formatter, enforce it with a Maven goal and you're done. The IntelliJ plugin does everything for you with Ctrl+Alt+L.

Yeah, that is indeed very radical change, it is even changing the line length to 100.

ibauersachs:
> Yeah, that is indeed very radical change, it is even changing the line length to 100.

So what?

(The Kotlin code uses longer lines as well).

damencho:
Its a step backwards from 120.

bgrozev (approved):

PR #240, opened 2021-11-08
Remove private jdk api

ibauersachs:


bgrozev (dismissed):

ibauersachs:
Thanks and sorry @bgrozev, but this isn't actually ready yet. One of the tests hangs reproducible when run on JDK17.

ibauersachs:
Maybe you've got an idea: when `org.ice4j.pseudotcp.PseudoTcpSocketImpl#close` is called in the client thread in `org.ice4j.pseudotcp.PseudoTcpStreamTest#testConnectTransferClose`, there's data left to be sent:
```
ON CLOSE: in flight 1334
ON CLOSE: buff not sent 555
```
That never happens on JDK8 or 11.

ibauersachs:
oO. So, `org.ice4j.pseudotcp.PseudoTcpSocketImpl.PseudoTcpOutputStream#flush` doesn't get called anymore because Java 17's new `SocketOutputStream` wraps `socketImpl.getOutputStream()`, but doesn't forward/delegate `flush()`.

bgrozev:
Hm, is the hanging test introduced in this PR or just exposed due to tests running on 17? If it's not a regression we can just add some comments and disable the test on 17

ibauersachs:
I think it is actually a regression since it's not possible anymore to flush the PseudoTcpOutputStream. However, I'm unsure if that flush is actually needed outside of tests.

bgrozev (dismissed):

ibauersachs:
So I think I've stabilized this enough to merge. Runs might still fail from time to time due to the randomness in the tests (which should be removed...).

bgrozev (approved):

nils-ohlmeier:
@ibauersachs is it possible that this PR causes problem with IPv4 vs IPv6 gathering. On my fresh test server ice4j appears to gather IPv6 only?!

PR #252, opened 2022-03-01
ref: Use the HEADER_LENGTH constant.

bgrozev:


JonathanLennox (approved):

src/main/java/org/ice4j/attribute/MessageIntegrityAttribute.java
```diff
@@ -313,7 +313,7 @@ else if (Message.isResponseType(msgType))
         this.hmacSha1Content = calculateHmacSha1(content, offset, length, key);
 
         //username
```
JonathanLennox, line 315: This comment is also wrong (pre-existing).

PR #259, opened 2022-05-02
ref: Remove use of JMap.

bgrozev:


JonathanLennox (approved):
Final commit should mention the removal of the Guava dependency, but otherwise lgtm.

PR #270, opened 2023-08-23
Randomize consent check interval (follow rfc7675).

bgrozev:


src/main/java/org/ice4j/ice/Agent.java
```diff
@@ -129,6 +129,8 @@ public void run()
     private final Map<String, IceMediaStream> mediaStreams
         = new LinkedHashMap<>();
 
+    SecureRandom random = new SecureRandom();
```
JonathanLennox, line 132: I don't think this needs SecureRandom; ThreadLocalRandom.current() should be fine.

JonathanLennox (approved):

PR #271, opened 2023-09-28
Update Kotlin to 1.9.10, and related Maven packages, and jitsi-utils and jicoco.

JonathanLennox:


pom.xml
```diff
@@ -38,10 +38,11 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <jicoco.version>1.1-104-g20c4353</jicoco.version>
-    <kotlin.version>1.6.21</kotlin.version>
-    <kotest.version>5.3.0</kotest.version>
-    <junit.version>5.8.2</junit.version>
+    <jicoco.version>1.1-126-g1f776be</jicoco.version>
```
bgrozev, line 41: I think you want 127 here to pick up the kotlin update?

JonathanLennox: You're right, thanks for catching that!

bgrozev (approved):

PR #274, opened 2023-11-14
fix: Delegate getInputStream of PseudoTcpSocket

fbieler:
Hello,

this is a small fix to allow writing data to pseudo TCP sockets. I presume it was an oversight that this method was not delegated.


bgrozev:
Re-open against [jitsi/jitsi-pseudotcp](https://github.com/jitsi/jitsi-pseudotcp/) is necessary (though as far as we know the project is not used)

PR #283, opened 2024-07-02
Add a config option to redact remote IP addresses in logs.

JonathanLennox:
And other human-readable text.

### DIFF
--- a/src/main/java/org/ice4j/TransportAddress.java
+++ b/src/main/java/org/ice4j/TransportAddress.java
@@ -161,6 +161,15 @@ public class TransportAddress
         if (redact)
         {
             hostAddress = getRedactedAddress();
+            if (hostAddress == null)
+            {
+                String hostName = getHostName();
+                if (hostName != null)
+                {
+                    /* The transport address is an unresolved hostname.  Redact the hostname. */
+                    hostAddress = "xxxx.xxx";
+                }
+            }
         }
         else
         {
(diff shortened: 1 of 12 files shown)


src/main/java/org/ice4j/TransportAddress.java
```diff
@@ -181,6 +199,19 @@ public String getHostAddress()
...
+    /* Return the host address, redacted if address redaction is enabled. */
+    public String getRedactedAddress()
+    {
+        if (AgentConfig.config.getRedactRemoteAddresses())
+        {
+            return toRedactedString(getAddress());
+        }
+        else
+        {
+            return getHostAddress();
```
bgrozev, line 211: This is missing the fallback to `getHostName()`, so changes behavior (though I don't understand the fallback).

src/main/java/org/ice4j/ice/Component.java
```diff
@@ -599,13 +599,22 @@ public String toString()
             buff.append("\n")
                 .append(remoteCandidatesCount)
                 .append(" Remote candidates:");
-            buff.append("\ndefault remote candidate: ")
-                .append(getDefaultRemoteCandidate());
+            Candidate<?> defaultRemoteCandidate = getDefaultRemoteCandidate();
+            buff.append("\ndefault remote candidate: ");
+            if (defaultRemoteCandidate != null)
+            {
+                buff.append(getDefaultRemoteCandidate().toRedactedString());
```
bgrozev, line 606: use the local `defaultRemoteCandidate` variable

src/main/java/org/ice4j/ice/Component.java
```diff
@@ -599,13 +599,22 @@ public String toString()
...
+            }
+            else
+            {
+                buff.append((String)null);
```
bgrozev, line 610: I think `append("null")` would be clearer, I wasn't sure what it does with an actuall `null`

bgrozev (approved):

PR #288, opened 2024-10-01
Run a first ICE connectivity check immediately, with no delay.

JonathanLennox:


src/main/java/org/ice4j/ice/ConnectivityCheckClient.java
```diff
@@ -972,6 +979,13 @@ protected void run()
          */
         protected Duration getDelayUntilNextRun()
```
bgrozev, line 980: Is it better to do this in `schedule`?

JonathanLennox: `schedule` is in the base class and there are other users of `PeriodicRunnable.`  We could have a `firstRunImmediate` flag in the base class which the subclass sets but it seemed unnecessarily complicated?

bgrozev (approved):

PR #293, opened 2024-10-15
Create a socket pool of multiple sockets bound to the same port

JonathanLennox:
So we can send from the same local IP and port from multiple threads simultaneously on Java 18+.


src/main/kotlin/org/ice4j/socket/SocketPool.kt
```diff
@@ -0,0 +1,114 @@
...
+        val socket: DatagramSocket,
+        var count: Int = 0
+    )
+
+    val numSockets: Int =
+        if (requestedNumSockets != 0) {
+            requestedNumSockets
+        } else {
+            // TODO: set this to 1 in situations where pools aren't needed?
+            2 * Runtime.getRuntime().availableProcessors()
```
bgrozev, line 60: I suggest logging the number we end up using.

Why does it default to twice the number of processors? Running your test on an A2 with 8 CPUs I see optimal performance for 8 sockets.

JonathanLennox: I was testing on a 4-core machine, but if JVB is using an 8 core one your tests are probably more representative.  (My tests were also pretty noisy; I suspect the host my lonely instance is on is pretty busy.)

bgrozev (dismissed):

bgrozev (approved):

PR #294, opened 2024-10-28
Add IMDSv2 support to AwsCandidateHarvester

dsmeytis:
Hello, AWS recommend to disable IMDSv1 support for EC2 instances: https://aws.amazon.com/blogs/security/get-the-full-benefits-of-imdsv2-and-disable-imdsv1-across-your-aws-infrastructure/
However it breaks AwsCandidateHarvester functionality because REST endpoints it uses to detect EC2 and obtain private/public IPs become unavailable. I would like to propose the fix that will work for both IMDSv1 and IMDSv2:
- `doTestEc2()` will now try to obtain an API token as `EC2_TEST_URL` is unavailable without a token.
- added parameters to `fetch()` to set a method and pass required headers.

### DIFF
--- a/src/main/java/org/ice4j/ice/harvest/AwsCandidateHarvester.java
+++ b/src/main/java/org/ice4j/ice/harvest/AwsCandidateHarvester.java
@@ -21,6 +21,8 @@ import org.ice4j.*;
 
 import java.io.*;
 import java.net.*;
+import java.util.Collections;
+import java.util.Map;
 import java.util.logging.*;
 
 /**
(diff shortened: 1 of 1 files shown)


dsmeytis:
already signed

src/main/java/org/ice4j/ice/harvest/AwsCandidateHarvester.java
```diff
@@ -183,11 +206,10 @@ private static boolean doTestEc2()
     {
         try
         {
-            URLConnection conn = new URL(EC2_TEST_URL).openConnection();
-            conn.setConnectTimeout(500); //don't hang for too long
```
bgrozev, line 187: Can you bring back the short timeout please?

src/main/java/org/ice4j/ice/harvest/AwsCandidateHarvester.java
```diff
@@ -21,6 +21,8 @@
 
 import java.io.*;
 import java.net.*;
+import java.util.Collections;
```
bgrozev, line 24: Change this to a `*` import please

bgrozev:
Thanks for the contribution! I left a couple of minor requests. I'll run a manual test and let you know if I have more.

Confirmed CLA. Jenkins, please add to whitelist

dsmeytis:
hi @bgrozev, thanks for the review! Actually for my deployment I ended up with the `java.net.http.HttpClient` implementation. If you wish I can push updated solution, otherwise I'll fix your requests and we'll proceed with it.

bgrozev:
Sure, unless it's a large change let's go with HttpClient

dsmeytis:
@bgrozev please take a look at https://github.com/jitsi/ice4j/pull/297

PR #297, opened 2024-12-03
Add IMDSv2 support to AwsCandidateHarvester and replace URLConnection…

dsmeytis:
… with HttpRequest


bgrozev:
Can we set a short timeout so it doesn't hang for a long time when it runs outside AWS?

dsmeytis:
@bgrozev added

bgrozev (approved):

bgrozev:
Thank you!

I just noticed that OCI also supports IMDSv2 and that address responds, so it ends up with RUNNING_ON_EC2=true. But that's unrelated to the PR.

PR #298, opened 2024-12-04
fix: Fail early if IMDS returns an error.

bgrozev:
This catches errors earlier on OCI for example, where the PUT request
returns 405.


bgrozev:
FYI @dsmeytis

